PR #58, opened 2026-04-09
feat: access control & authorisation layer

### LINKED ISSUE #38: Implement: Access Control & Security — Shillinq
## OpenSpec Change: access-control-authorisation

Spec path: `openspec/changes/access-control-authorisation`

rubenvdlinde:
Closes #38

## Summary
Implements the full access control and authorisation subsystem for Shillinq. This adds role-based access control with hierarchical levels (Admin/Editor/Accountant/Viewer/Reports-only), field-level security, purchasing limits, time-limited delegation, access recertification campaigns, and an immutable audit log. Both backend (PHP middleware, services, controllers, background jobs) and frontend (Pinia stores, Vue views, navigation, i18n) are included.

## Spec Reference
- Issue: #38
- Spec: `openspec/changes/access-control-authorisation/design.md`

## Changes
- `lib/Settings/shillinq_register.json` — Replaced example schema with 7 new schemas: role, team, user, accessControl, accessRight, accessRecertification, permission
- `lib/Repair/CreateDefaultConfiguration.php` — Seeds 5 built-in roles, 1 team, and 3 sample audit events
- `lib/Middleware/PermissionGateMiddleware.php` — Enforces role-level checks and isActive gating on all Shillinq API requests
- `lib/Middleware/PermissionDeniedException.php` — Custom exception for permission denials
- `lib/Service/AuditLogService.php` — Fire-and-forget audit logging via AccessControl objects
- `lib/Service/FieldSecurityService.php` — Field-level read/write enforcement based on permission matrix
- `lib/Service/PurchasingLimitService.php` — Per-buyer purchasing limit checks across base + delegated roles
- `lib/Service/DelegationService.php` — Create/revoke time-limited AccessRight delegations with notifications
- `lib/Service/RecertificationService.php` — Dispatch review notifications and process confirm/revoke decisions
- `lib/BackgroundJob/DelegationExpiryJob.php` — Auto-revokes expired delegations every 5 minutes
- `lib/BackgroundJob/RecertificationNotificationJob.php` — Hourly cron-based recertification campaign dispatch
- `lib/Controller/RoleController.php` — CRUD for roles with delete-protection for assigned users
- `lib/Controller/TeamController.php` — CRUD for teams with member invite endpoint
- `lib/Controller/UserController.php` — List/show/update users and HR provisioning endpoint
- `lib/Controller/AccessControlController.php` — Read-only audit log with action/result filters
- `lib/Controller/DelegationController.php` — Create/revoke delegations
- `lib/Controller/RecertificationController.php` — Campaign management and review decision submission
- `lib/Controller/ReportController.php` — Access rights report in JSON or CSV format
- `lib/AppInfo/Application.php` — Registered middleware and repair step
- `appinfo/routes.php` — Added all new API routes
- `src/store/modules/role.js` — Pinia store for roles
- `src/store/modules/team.js` — Pinia store for teams
- `src/store/modules/user.js` — Pinia store for users
- `src/store/modules/accessControl.js` — Pinia store for audit events
- `src/store/modules/delegation.js` — Pinia store for delegations
- `src/store/modules/recertification.js` — Pinia store for recertification campaigns
- `src/store/store.js` — Registered all 6 new stores
- `src/views/role/` — RoleIndex and RoleDetail views
- `src/views/team/` — TeamIndex, TeamDetail, and TeamInviteDialog views
- `src/views/user/` — UserIndex and UserDetail views
- `src/views/accessControl/` — AccessControlIndex and AccessControlDetail views
- `src/views/delegation/DelegationDialog.vue` — Delegation creation dialog
- `src/views/recertification/` — RecertificationIndex and RecertificationReview views
- `src/views/report/AccessRightsReport.vue` — Access rights report with CSV export
- `src/navigation/MainMenu.vue` — Added Security section with 6 nav items
- `src/router/index.js` — Added all new routes with breadcrumb metadata
- `l10n/en.json` — English translations for all access control UI strings
- `l10n/nl.json` — Dutch translations

## Test Coverage
- `tests/unit/Middleware/PermissionGateMiddlewareTest.php` — Inactive user denial, 403 JSONResponse, non-gated controller passthrough
- `tests/unit/Service/FieldSecurityServiceTest.php` — Restricted field stripping, admin access, permission cache
- `tests/unit/Service/PurchasingLimitServiceTest.php` — Over-limit block, category mismatch, higher delegated limit
- `tests/unit/BackgroundJob/DelegationExpiryJobTest.php` — Expired delegation revocation, future delegation unchanged, audit event on expiry


rubenvdlinde:

Fixed PHPUnit failures in `feature/38/access-control-authorisation`:

- **[FAILURE] `FieldSecurityServiceTest::testFilterResponseStripsRestrictedFields`** — `bankAccountNumber` was not stripped from the response.
- **[FAILURE] `PurchasingLimitServiceTest::testAmountExceedingLimitReturnsBlocked`** — limit check returned \`true\` (allowed) instead of \`false\` (blocked).

**Root causes (both services shared the same two bugs):**

1. **Named arguments on PHPUnit `addMethods()` mocks** — calling \`findObjects(filters: X, register: Y, schema: Z)\` on a mock generated with \`addMethods()\` throws an \`Error\` because the generated method has no declared parameter names. The \`catch(\\Throwable)\` silently swallowed the error, returning fail-open values that masked the real problem.

2. **`Application::APP_ID` triggers OCP autoloading** — referencing \`Application::APP_ID\` causes PHP to autoload \`OCA\\Shillinq\\AppInfo\\Application\`, which extends \`OCP\\AppFramework\\App\`. Outside the Nextcloud environment that class does not exist, so every \`findObjects()\` call threw before executing any logic.

**Fix (commit `30b6c14`):**
- Changed all \`findObjects()\` calls from named to positional arguments.
- Replaced \`Application::APP_ID\` with the inline string \`'shillinq'\` and removed the now-unused \`use\` import.

**Quality gate results after fix:**
- PHPCS: ✅ 25/25 files pass
- PHPUnit: ✅ 0 failures (9 errors remain — all OCP class-not-found, expected outside Nextcloud env)

rubenvdlinde:
## Code Review — Juan Claude van Damme (Hydra)

**PR:** feat: access control & authorisation layer (#58)
**Reviewer:** Juan Claude van Damme — Hydra Code Review Agent
**Date:** 2026-04-09

---

### Overview

Reviewed all 115 changed files. The access control architecture is well-structured overall — the middleware/service layering follows the spec, EUPL-1.2 headers are present on all new PHP files, and the test coverage for the four unit-tested classes is solid. However, there are **four blocking issues** that must be addressed before merge.

rubenvdlinde:
**[CRITICAL] Background jobs never registered — delegations will never auto-expire**
`lib/AppInfo/Application.php:60-77`
Issue: `DelegationExpiryJob` and `RecertificationNotificationJob` are fully implemented but never registered with `$context->registerBackgroundJob()` in `Application::register()`. The Nextcloud job scheduler has no knowledge of them, so they will never run. This means time-limited delegations will remain active indefinitely after their `endDate`, and access recertification campaigns will never trigger — both core requirements of the spec.
Fix: Add the two registration calls inside `register()`:
```php
use OCA\Shillinq\BackgroundJob\DelegationExpiryJob;
use OCA\Shillinq\BackgroundJob\RecertificationNotificationJob;

// inside register():
$context->registerBackgroundJob(DelegationExpiryJob::class);
$context->registerBackgroundJob(RecertificationNotificationJob::class);
```

rubenvdlinde:
**[CRITICAL] Middleware allows all Nextcloud users without a Shillinq profile to bypass access control**
`lib/Middleware/PermissionGateMiddleware.php:107-112`
Issue: When `resolveUser()` returns `null` (i.e., the authenticated Nextcloud user has no Shillinq User object), the middleware logs a `success` event and **returns without any role check**. This gives every valid Nextcloud user — including those who were never provisioned — unrestricted access to all gated endpoints (RoleController, DelegationController, UserController, etc.). This completely negates the RBAC enforcement for the initial install state and for any user who was not explicitly provisioned.
Fix: Treat a missing Shillinq profile as an access denial rather than a pass-through:
```php
if ($shillinqUser === null) {
    $this->auditLogService->log('permission-denied', 'api', null, 'denied', ['reason' => 'no shillinq profile']);
    throw new PermissionDeniedException('No Shillinq profile found for this user');
}
```
If certain endpoints must remain open before provisioning, apply a dedicated allow-list annotation (e.g., `@NoPermissionGate`) rather than silently passing through.

rubenvdlinde:
**[CRITICAL] `grantedBy` in delegation creation is caller-controlled — audit log non-repudiation is broken**
`lib/Controller/DelegationController.php:71`
Issue: The `grantedBy` field — the identity of the administrator authorising the delegation — is taken directly from the request body (`$this->request->getParam('grantedBy', '')`). Any user with role level 100 can pass an arbitrary user ID as `grantedBy`, attributing their actions to someone else in the immutable audit log. This undermines the entire audit trail, which is a core compliance requirement of the access control subsystem.
Fix: Resolve `grantedBy` from the authenticated session on the server side — never trust the client:
```php
// Inject IUserSession via constructor, then:
$ncUser = $this->userSession->getUser();
$shillinqUsers = $objectService->findObjects(['username' => $ncUser->getUID()], ...);
$grantedBy = $shillinqUsers[0]['id'] ?? '';
```
Remove `grantedBy` from the accepted request parameters entirely.

rubenvdlinde:
**[CRITICAL] All Vue views use `@nextcloud/vue` — architecture requires `@conduction/nextcloud-vue` components**
`src/views/role/RoleIndex.vue:71`, `src/views/user/UserDetail.vue:78`, and all other view files
Issue: The Conduction architecture spec (CLAUDE.md, `openspec/changes/access-control-authorisation/design.md`) mandates use of `@conduction/nextcloud-vue` components — specifically `CnIndexPage`, `CnDetailPage`, and `CnFormDialog` — for all index/detail/dialog page patterns. All 13 new Vue view files instead import directly from `@nextcloud/vue` (`NcButton`, `NcDialog`, `NcBadge`, etc.) and implement custom table/tab/breadcrumb layouts by hand. This is a direct spec deviation and will require a full frontend rewrite to bring into compliance.
Fix: Replace all custom table/tab/dialog implementations with the appropriate `@conduction/nextcloud-vue` composites:
- List pages → `CnIndexPage` with `columnsFromSchema` / `filtersFromSchema`
- Detail pages → `CnDetailPage` with tab slots
- Dialogs → `CnFormDialog` with `fieldsFromSchema`

See `design.md §Frontend Components` for the prescribed pattern per view.

rubenvdlinde:
**[WARNING] Purchasing limit service fails open on exception — dangerous for a financial control**
`lib/Service/PurchasingLimitService.php:130-133`
Issue: The entire `checkLimit()` body is wrapped in a try/catch that returns `true` (allow purchase) on any `\Throwable`. If the OpenRegister service is unavailable, the database is slow, or any unexpected error occurs, the purchasing limit check silently passes. For a financial access control this is the wrong failure mode — a flaky infrastructure condition becomes an open door for over-limit purchases.
Fix: Fail closed. Return `false` on error, and surface the failure clearly:
```php
} catch (\Throwable $e) {
    $this->logger->error('Shillinq: failed to check purchasing limit — denying as fail-safe', ['exception' => $e->getMessage()]);
    return false;
}
```
If a fail-open behaviour is a deliberate business decision, document it explicitly in the spec and add a log entry at ERROR level (currently it logs at ERROR but still returns `true`).

rubenvdlinde:
**[WARNING] CSV injection — formula-starting characters not sanitised in report export**
`lib/Controller/ReportController.php:148-158`
Issue: `generateCsv()` escapes commas and double-quotes but does not sanitise values that begin with `=`, `+`, `-`, or `@`. When the CSV is opened in Excel or LibreOffice Calc, these characters trigger formula evaluation. A user `displayName` or `branch` value of `=CMD|' /C calc'!A0` would execute arbitrary commands on the reviewer's machine. Given that this is an access-rights export likely opened by compliance officers, this is a real attack surface.
Fix: Prefix any cell value that starts with `=`, `+`, `-`, or `@` with a tab or single-quote:
```php
if (preg_match('/^[=+\-@]/', $value)) {
    $value = "\t" . $value;
}
```
Apply this before the existing comma/quote escaping.

rubenvdlinde:
**[WARNING] DelegationController does not validate date inputs — malformed dates cause unhandled 500**
`lib/Controller/DelegationController.php:85-87`
Issue: `new \DateTime($startDate)` and `new \DateTime($endDate)` are called on whatever strings the client provides. Passing an empty string (`''`) or an invalid date (`'not-a-date'`) causes `\DateTime` to throw a generic `\Exception`, which is not caught (`DelegationService` only throws `\InvalidArgumentException`). This results in an unhandled 500 response rather than a meaningful 422.
Fix: Validate the date strings before constructing `\DateTime` objects, or wrap the construction in a try/catch at the controller level:
```php
try {
    $start = new \DateTime($startDate);
    $end   = new \DateTime($endDate);
} catch (\Exception $e) {
    return new JSONResponse(['error' => 'Invalid date format for startDate or endDate'], 422);
}
```

rubenvdlinde:
**[WARNING] FieldSecurityService permission cache is not keyed by user ID**
`lib/Service/FieldSecurityService.php:127-129`
Issue: `loadPermissions()` populates `$this->permissionCache` on the first call, then returns the cached value unconditionally on all subsequent calls regardless of `$userId`. If the service instance is shared (it is a DI singleton) and `filterResponse()` or `checkWritePermission()` is called for different users within the same request lifecycle (e.g., via a bulk endpoint or a future service-to-service call), user B gets user A's permission set.
Fix: Key the cache by user ID:
```php
private array $permissionCache = [];

private function loadPermissions(string $userId): array
{
    if (isset($this->permissionCache[$userId])) {
        return $this->permissionCache[$userId];
    }
    // ... load ...
    $this->permissionCache[$userId] = $allPermissions;
    return $this->permissionCache[$userId];
}
```

rubenvdlinde:
**[WARNING] UserDetail tabs "Roles & Permissions", "Delegations", and "Access History" are empty stubs**
`src/views/user/UserDetail.vue:55-68`
Issue: Three of the four tabs in `UserDetail.vue` are placeholder paragraphs with no data rendering. The spec (`design.md §Frontend Components`) explicitly requires the Roles tab to list assigned roles with role badge, the Delegations tab to list active `AccessRight` objects, and the Access History tab to show the last 50 `AccessControl` events filtered by `userId`. These are visible to end users and must be implemented before the feature is considered complete.
Fix: Implement each tab body as described in the design spec:
- Roles tab: call `accessRightStore.fetchForUser(userId)` and render role badges
- Delegations tab: render active delegations from the same store, use `DelegationDialog` for creation
- Access History tab: call `accessControlStore.fetchForUser(userId, { limit: 50 })` and render a read-only table

rubenvdlinde:
**[SUGGESTION] UserController.provision() hard-codes `@example.com` as email domain**
`lib/Controller/UserController.php:163`
Issue: Provisioned users receive `$employeeId . '@example.com'` as their email address. Any delegation notification or recertification email dispatched to this user will go to a nonexistent address. This is likely a placeholder that was never replaced.
Fix: Either require `email` as a parameter in the provision endpoint, or derive it from Nextcloud's user manager (`IUserManager::get($employeeId)->getEMailAddress()`):
```php
$email = $this->request->getParam('email', '');
if (empty($email)) {
    $ncUser = $this->userManager->get($employeeId);
    $email  = $ncUser !== null ? $ncUser->getEMailAddress() : '';
}
```

rubenvdlinde:
```json
{ "pass": false, "blocking": ["Background jobs DelegationExpiryJob and RecertificationNotificationJob are not registered in Application.php — delegations will never auto-expire", "PermissionGateMiddleware allows all Nextcloud users without a Shillinq profile to bypass RBAC on all gated endpoints", "DelegationController accepts grantedBy from the request body — audit log non-repudiation is broken", "All Vue views use @nextcloud/vue components instead of the required @conduction/nextcloud-vue CnIndexPage/CnDetailPage/CnFormDialog components", "PurchasingLimitService fails open on exception — financial control must fail closed", "ReportController CSV export is vulnerable to formula injection via unsanitised = + - @ prefixes"] }
```

rubenvdlinde:
## Hydra Security Review — Clyde Barcode

**Scope:** `lib/Service/FieldSecurityService.php`, `lib/Service/PurchasingLimitService.php`
**SAST:** Semgrep `p/security-audit` + `p/secrets` + `p/owasp-top-ten` — 0 automated findings
**Manual review:** OWASP Top 10:2025 + Conduction security rules

---

### Finding 1

**[CRITICAL] Permission cache not keyed by userId — horizontal privilege escalation**
Rule: OWASP A01:2021 Broken Access Control
`lib/Service/FieldSecurityService.php:127`
Issue: `loadPermissions()` caches the first user's permission set in `$this->permissionCache` and returns it for **every subsequent call regardless of the `$userId` argument**. If the service instance is shared across multiple users in a single PHP lifecycle (batch job, queue worker, CLI command, or any code path calling `filterResponse()` / `checkWritePermission()` for more than one user), the first user's permissions are applied to all users. A low-privilege user processed first would cause subsequent higher-privilege users to have fields stripped; more critically a high-privilege (or no-restriction) user processed first would expose all restricted fields to a low-privilege user processed next — a horizontal access-control bypass.
Fix: Key the cache by userId:
```php
private array $permissionCache = [];

private function loadPermissions(string $userId): array
{
    if (isset($this->permissionCache[$userId])) {
        return $this->permissionCache[$userId];
    }
    // … existing load logic …
    $this->permissionCache[$userId] = $allPermissions;
    return $this->permissionCache[$userId];
}
```

---

### Finding 2

**[WARNING] Fail-open on exception in a financial enforcement control**
Rule: OWASP A04:2021 Insecure Design
`lib/Service/PurchasingLimitService.php:130-133`
Issue: When `ObjectService` throws (service unavailable, timeout, unexpected data), the catch block logs and **returns `true` (purchase allowed)**. An attacker who can trigger repeated ObjectService failures — or a legitimate outage — causes all purchasing limit checks to be silently bypassed. The comment acknowledges this is intentional (`// Fail open: allow the purchase if the check fails`), but for a financial enforcement control this is an insecure default.
Fix: Fail closed — throw a typed exception that the caller must handle, or return `false`. If fail-open is a product requirement, document it as an accepted risk at the ADR level and alert/monitor for check failures rather than silently passing.

---

### Finding 3

**[WARNING] No active roles → purchase unconditionally allowed**
Rule: OWASP A01:2021 Broken Access Control
`lib/Service/PurchasingLimitService.php:85-87`
Issue: When a user has zero active access rights, `checkLimit()` returns `true` with no further evaluation. A user whose access rights have been revoked, or a newly provisioned user with no role assignment, can make unlimited purchases until roles are assigned. This may be intended, but it is a default-allow posture that warrants an explicit product decision and documentation.
Fix: Return `false` (blocked) when no roles are found, unless product requirements explicitly mandate unrestricted purchasing for un-roled users. Document the decision in the relevant ADR.

---

### Finding 4

**[SUGGESTION] Deny-list field enforcement relies on explicit `canRead: false` — missing entries are implicitly allowed**
Rule: OWASP A01:2021 — defence-in-depth
`lib/Service/FieldSecurityService.php:80-82`
Issue: `filterResponse()` only removes a field when a permission entry with `canRead === false` exists. Fields with no permission entry at all pass through. This deny-list model means any new schema field is readable by all users until an explicit deny is authored. An allow-list model (only include fields with `canRead === true`, strip everything else) is safer for sensitive data schemas like `supplier`.
Fix: Consider inverting the model for sensitive schemas: build an explicit allow-set of fields and strip everything not in it.

---

### SAST

[FALSE POSITIVE] — No Semgrep findings were generated. 63 rules across `p/security-audit`, `p/secrets`, and `p/owasp-top-ten` ran clean against both files. No false positives to suppress.

rubenvdlinde:
```json
{ "pass": false, "blocking": ["Permission cache not keyed by userId in lib/Service/FieldSecurityService.php:127"] }
```

rubenvdlinde:
## Phase 1 — Context & Overview

**PR:** feat: access control & authorisation layer (#58)
**Size:** 6 545 additions / 64 deletions across 50 files — this is a large PR but the scope (a complete RBAC subsystem) warrants it.
**Spec:** openspec/changes/access-control-authorisation/design.md — read and cross-referenced.
**CI status:** PR is in DRAFT; no CI results yet.

I reviewed all changed PHP files (middleware, services, controllers, background jobs, tests) and representative Vue/JS files. Findings are posted individually below, then a final verdict.

rubenvdlinde:
**[CRITICAL] Audit log entries omit the actor (who performed the action)**
`lib/Service/AuditLogService.php:76-95`
Issue: `AuditLogService::log()` saves action, resource, result, IP, and user-agent, but never records *who* performed the action. The spec data model for `AccessControl` explicitly lists a User actor relation: _"Relations: → User (actor), → AccessRight (if delegation-related)"_. An audit trail without an actor cannot answer "who did what" and fails any compliance or forensic review purpose.
Fix: Inject `IUserSession` into `AuditLogService` (it is already available in the Nextcloud DI container) and write `'actor' => $this->userSession->getUser()?->getUID()` into the $data array before calling `saveObject()`.

rubenvdlinde:
**[CRITICAL] All new frontend views bypass required `@conduction/nextcloud-vue` components**
`src/views/role/RoleIndex.vue`, `src/views/role/RoleDetail.vue`, `src/views/team/`, `src/views/user/`, `src/views/accessControl/`, `src/views/delegation/`, `src/views/recertification/`, `src/views/report/`
Issue: Existing views (Dashboard, Settings) correctly use `CnIndexPage`, `CnDetailPage`, `CnSettingsSection`, and `CnFormDialog` from `@conduction/nextcloud-vue`. All 14 new views use raw `<table>`, `<div class="..."\>`, and custom scoped `<style>` blocks instead, diverging from the established architecture pattern. The architecture spec (design.md) states: _"The Vue 2.7 + Pinia frontend handles all CRUD rendering via the standard `CnIndexPage` / `CnDetailPage` / `CnFormDialog` pattern."_
Fix: Rewrite all new views to use the `CnIndexPage` / `CnDetailPage` / `CnFormDialog` components. Remove the per-view custom CSS that duplicates design-system tokens already handled by the component library.

rubenvdlinde:
**[CRITICAL] `PurchasingLimitService::checkLimit()` fails open on exception**
`lib/Service/PurchasingLimitService.php:130-134`
Issue: When the ObjectService call throws (e.g. OpenRegister is unavailable, database error, network timeout), the catch block logs the error and returns `true` — **allowing** the purchase. The comment even documents this: _"Fail open: allow the purchase if the check fails."_ For a purchasing-limit enforcement service that guards financial transactions, fail-open is the wrong default. An attacker who can cause a transient ObjectService error gains unlimited purchasing authority.
Fix: Change the catch block to return `false` (deny), or at minimum surface the error as a 503 via the caller rather than silently permitting the transaction. Fail-closed is the correct default for a security enforcement path.

rubenvdlinde:
**[WARNING] N+1 query pattern in permission and report paths**
Files:
- `lib/Middleware/PermissionGateMiddleware.php:261-274` — one DB query per access-right to fetch the role
- `lib/Service/PurchasingLimitService.php:93-101` — one DB query per role ID inside a loop
- `lib/Service/FieldSecurityService.php:162-169` — one permission query per role ID inside a loop
- `lib/Controller/ReportController.php:86-104` — per-user × per-delegation × per-role nested queries

Issue: These patterns issue O(N) separate database/ObjectService calls where N is the number of access rights or users. For a system with 100 users and 5 roles each, the access-rights report alone makes 500+ round trips. Under load this will be a significant performance problem, and the middleware check runs on *every* gated API request.
Fix: Use a single `findObjects` call with an `id IN [...])` filter (or equivalent batch API) after collecting all role IDs, rather than querying each role individually inside a loop. Confirm with the OpenRegister `ObjectService` API whether it supports `id` as an array filter.

rubenvdlinde:
**[WARNING] Notification API receives Shillinq object ID instead of Nextcloud user UID**
`lib/Service/DelegationService.php:183`, `lib/BackgroundJob/DelegationExpiryJob.php:148`
Issue: `$notification->setUser($userId)` is called where `$userId` is a Shillinq `AccessRight.userId` — an OpenRegister object UUID, **not** a Nextcloud account UID (username). Nextcloud's notification manager requires a Nextcloud user ID for delivery. Sending a UUID will silently fail (the exception is caught by the surrounding `\Throwable` handler at DelegationService.php:186), meaning all delegation-created, delegation-revoked, and delegation-expired notifications are effectively dead on arrival.
`RecertificationService::dispatchReviewNotifications()` correctly uses `$user['username']`, which is the right field.
Fix: Resolve the Shillinq User object for the given `userId`, then call `setUser($user['username'])` to pass the Nextcloud username.

rubenvdlinde:
**[WARNING] `RecertificationService::processReviewDecisions()` silently accepts invalid actions as 'confirm'**
`lib/Service/RecertificationService.php:111-144`
Issue: The `foreach` over `$decisions` only checks `if ($decision['action'] === 'revoke')`. Any other value — including typos ('confirmed', 'approve') or malformed input — falls through to the `else` branch and increments `$confirmed`. A reviewer who sends `action: 'suspend'` would silently confirm the access right rather than get an error.
Fix: Validate that `action` is one of `['confirm', 'revoke']` before processing, and return an error (or skip with a log warning) for unrecognised values:
```php
if (!in_array($decision['action'], ['confirm', 'revoke'], true)) {
    $this->logger->warning('Unknown recertification action', ['action' => $decision['action']]);
    continue;
}
```

rubenvdlinde:
**[WARNING] Cron expression handling is an unreliable heuristic, not spec-compliant**
`lib/BackgroundJob/RecertificationNotificationJob.php:145-168`
Issue: `computeNextRun()` uses a three-branch heuristic (day-of-month set → +1 month, hour set → +1 day, else → +1 hour) to compute the next scheduled run from a cron string. The spec states that `cronExpression` is a standard cron string such as `0 9 1 * *`, implying proper cron semantics. The heuristic fails for many valid expressions:
- `0 9 * * 1` (every Monday) → incorrectly classified as "hourly" (parts[1]='9' is not '*' but parts[2]='*'), so adds 1 day — sends on consecutive days instead of weekly
- `0 9 1,15 * *` (1st and 15th) → parts[2]='1,15' ≠ '*', so adds 1 month — skips the 15th entirely
Fix: Add `dragonmantank/cron-expression` (already common in the PHP/Nextcloud ecosystem) to `composer.json` and replace the heuristic with `CronExpression::getNextRunDate()`.

rubenvdlinde:
**[WARNING] `FieldSecurityService` permission cache ignores the `userId` parameter**
`lib/Service/FieldSecurityService.php:127-128`
Issue: `loadPermissions($userId)` returns `$this->permissionCache` on the second call regardless of what `$userId` is. Within a single request this is safe (one user per request). However, if the service is ever injected into a context where it is called for two different users — e.g. a background job checking field permissions for multiple users in one run, or a future admin operation — the first user's permission set will be silently reused for all subsequent users.
Fix: Key the cache by userId:
```php
private array $permissionCache = [];

private function loadPermissions(string $userId): array {
    if (isset($this->permissionCache[$userId])) {
        return $this->permissionCache[$userId];
    }
    // ... load ...
    $this->permissionCache[$userId] = $allPermissions;
    return $this->permissionCache[$userId];
}
```

rubenvdlinde:
**[WARNING] `TeamController::invite()` derives username from email without validation**
`lib/Controller/TeamController.php:225`
Issue: `'username' => explode('@', $email)[0]` will produce an empty string if the email address starts with `@` (e.g. `@example.com`), or an incorrect value for quoted local-parts. The user object is then saved with `username: ''`, which makes the isActive-gate in `PermissionGateMiddleware::resolveUser()` unreliable for that user, and can cause duplicate-user issues since `findObjects(['email' => $email])` is used for deduplication but the username derivation happens unconditionally.
Fix: Add `filter_var($email, FILTER_VALIDATE_EMAIL)` before processing, and return a 422 if invalid. Derive the username from the Nextcloud user lookup rather than a string split — or at minimum assert the local-part is non-empty before saving.

rubenvdlinde:
**[SUGGESTION] Insufficient test coverage — controllers and key services untested**
Files: `tests/unit/`
Issue: The PR adds 4 unit test files covering: `PermissionGateMiddleware`, `FieldSecurityService`, `PurchasingLimitService`, `DelegationExpiryJob`. Not covered: `DelegationService` (delegation creation/revocation logic), `RecertificationService` (processReviewDecisions), `RecertificationNotificationJob` (cron scheduling), and all 7 controllers (RoleController, TeamController, UserController, AccessControlController, DelegationController, RecertificationController, ReportController). The report CSV generation path in particular has no test despite having logic (N+1 loop, CSV escaping).
Fix: Add unit tests for at minimum `DelegationService::createDelegation()`/`revokeDelegation()`, `RecertificationService::processReviewDecisions()` (especially the unknown-action edge case), and `ReportController::accessRights()` in both JSON and CSV formats.

rubenvdlinde:
**[SUGGESTION] CSV export is vulnerable to formula injection**
`lib/Controller/ReportController.php:134-167`
Issue: `generateCsv()` escapes commas and double-quotes but does not sanitise values that begin with `=`, `+`, `-`, or `@`. A Shillinq user whose `username` or `displayName` starts with one of these characters (e.g. `=cmd|'/C calc'!A1`) can inject spreadsheet formulas that execute when the exported CSV is opened in Excel or LibreOffice Calc. While this is an internal tool, the CSV is downloaded directly by admin users and may be opened without stripping.
Fix: Prefix any cell value starting with `=`, `+`, `-`, or `@` with a single quote or a tab character before output, or wrap all values in double quotes unconditionally (which also simplifies the escaping logic).

rubenvdlinde:
**[SUGGESTION] `RoleDetail.vue` Permissions and Members tabs are stub placeholders**
`src/views/role/RoleDetail.vue:49-55`
Issue: Both the "Permissions" and "Members" tabs render only a single static paragraph and no data. If these are intentionally deferred to a follow-up PR, please add a `// TODO` comment or link to the tracking issue so reviewers can distinguish intentional stubs from forgotten implementations.
Fix: Either implement the tabs or add a spec link/issue reference in the placeholder comment.

rubenvdlinde:
```json
{ "pass": false, "blocking": ["[CRITICAL] AuditLogService::log() never records the actor (who performed the action) — audit trail is non-compliant with spec AccessControl data model", "[CRITICAL] All 14 new frontend views bypass the required CnIndexPage/CnDetailPage/CnFormDialog component pattern from @conduction/nextcloud-vue — architecture compliance violation", "[CRITICAL] PurchasingLimitService::checkLimit() fails open on ObjectService exception, silently allowing purchases above the authorised limit", "[WARNING] N+1 query pattern in PermissionGateMiddleware::getUserRoleLevel(), PurchasingLimitService::checkLimit(), FieldSecurityService::loadPermissions(), and ReportController::accessRights()", "[WARNING] DelegationService and DelegationExpiryJob pass Shillinq object UUID to INotificationManager::setUser() instead of Nextcloud user UID — all delegation notifications are silently dropped", "[WARNING] RecertificationService::processReviewDecisions() treats any action other than \"revoke\" as a confirmed decision, including invalid/typo values", "[WARNING] RecertificationNotificationJob::computeNextRun() uses an ad-hoc heuristic instead of a proper cron parser, producing incorrect schedules for weekly and bi-monthly expressions", "[WARNING] FieldSecurityService permission cache is keyed by instance (not userId), causing incorrect permission reuse if called for multiple users in one lifecycle", "[WARNING] TeamController::invite() derives username from email without validation — empty username saved when email starts with @"] }
```

rubenvdlinde:
## Security Review — FieldSecurityService.php + PurchasingLimitService.php

**Reviewer:** Clyde Barcode (Hydra Security Reviewer, Conduction B.V.)
**Scan:** Semgrep OSS 1.135.0 — rulesets `p/security-audit`, `p/secrets`, `p/owasp-top-ten` (63 rules, 0 automated findings)
**Manual review:** OWASP Top 10:2025 + Conduction ADR-005

---

### Semgrep Results
✅ 0 automated findings across both files (p/security-audit, p/secrets, p/owasp-top-ten).

---

### Manual OWASP Review Findings

---

**[CRITICAL] Permission cache not keyed by userId — cross-user permission bleed**
Rule: OWASP A01:2025 – Broken Access Control
`lib/Service/FieldSecurityService.php:127`
Issue: `$permissionCache` is a plain instance-level property with no user-ID key. The first call to `loadPermissions()` populates it; every subsequent call returns that cached result regardless of the `$userId` argument. In Nextcloud, DI-injected services are typically shared singletons within a request. Any code path that calls `filterResponse()` or `checkWritePermission()` for more than one user in the same request (admin operations, batch jobs, event listeners) will silently evaluate the second user's permissions against the first user's permission set. This can result in data disclosure (fields that should be stripped are not) or privilege escalation (writes that should be blocked are permitted).
Fix: Key the cache by userId:
```php
private array $permissionCache = [];

private function loadPermissions(string $userId): array
{
    if (array_key_exists($userId, $this->permissionCache)) {
        return $this->permissionCache[$userId];
    }
    // ... load ...
    $this->permissionCache[$userId] = $allPermissions;
    return $this->permissionCache[$userId];
}
```

---

**[CRITICAL] Fail-open on exception in a financial enforcement control**
Rule: OWASP A10:2025 – Mishandling of Exceptional Conditions
`lib/Service/PurchasingLimitService.php:130-133`
Issue: The catch block explicitly returns `true` (purchase allowed) when `ObjectService` throws. The code comment confirms this is intentional: _"Fail open: allow the purchase if the check fails."_ For a financial or compliance enforcement boundary, failing open is dangerous: an attacker who can trigger transient errors in `ObjectService` (connection exhaustion, forced timeouts, crafted input that causes an exception) fully bypasses all purchasing limits.
Fix: Fail closed — return `false` and let the caller surface an error to the user:
```php
} catch (\Throwable $e) {
    $this->logger->error('Shillinq: failed to check purchasing limit', ['exception' => $e->getMessage()]);
    return false; // fail closed — deny on error
}
```

---

**[WARNING] Fail-open default: users with no roles pass all field-security and limit checks**
Rule: OWASP A06:2025 – Insecure Design / OWASP A01:2025 – Broken Access Control
Files: `lib/Service/FieldSecurityService.php:141-143` · `lib/Service/PurchasingLimitService.php:85-87`
Issue: Both services treat an empty role set as "no restrictions":
- `FieldSecurityService`: when a user has no access rights, `loadPermissions()` returns `[]`, no fields are stripped, and `checkWritePermission()` returns `true`. The user can read and write all fields.
- `PurchasingLimitService`: when `$roleIds` is empty the method immediately returns `true` — unlimited purchasing.
A newly provisioned user, an account with all expired delegations, or any account with missing `accessRight` records gets full access rather than zero access. The same fail-open applies on exception in `loadPermissions()` (line 173-176).
Fix: Adopt a default-deny stance. If no roles are resolved, return the most restrictive result until explicit permissions are granted.

---

**[WARNING] Negative `$amount` unconditionally bypasses purchasing limit**
Rule: OWASP A05:2025 – Input Validation
`lib/Service/PurchasingLimitService.php:129`
Issue: `$amount` is typed `float` but never validated to be ≥ 0. The comparison `$amount <= $highestLimit` is always `true` for negative values, so any caller that passes a user-influenced negative amount bypasses the limit entirely. Depending on downstream handling, negative amounts may also produce negative spend records that offset against legitimate spending.
Fix: Reject non-positive amounts at the top of `checkLimit()`:
```php
if ($amount < 0) {
    $this->logger->warning('Shillinq: negative purchase amount rejected', ['userId' => $userId, 'amount' => $amount]);
    return false;
}
```

rubenvdlinde:
## Juan Claude van Damme — Code Review

**PR:** feat: access control & authorisation layer (#58)
**Reviewer:** Juan Claude van Damme (Hydra CI/CD)
**Date:** 2026-04-09

---

This is a large, well-structured PR (6 545 additions) implementing the full RBAC subsystem. The backend design is clean and the test coverage for the included unit tests is solid. However, I found three **CRITICAL** and six **WARNING** items that must be addressed before merge.

---

### Finding 1

**[CRITICAL] Fail-open bypass: gated endpoints reachable with no Shillinq user profile**
`lib/Middleware/PermissionGateMiddleware.php:97-101`
Issue: When `resolveUser()` returns `null` (either because OpenRegister is unavailable OR because the authenticated Nextcloud user has no Shillinq `User` object), the middleware immediately returns without checking the `$gatedControllers` list. This means every gated endpoint — `RoleController`, `TeamController`, `UserController`, `AccessControlController`, `DelegationController`, `RecertificationController`, `ReportController` — is **fully open** to any authenticated Nextcloud user who has not been provisioned in Shillinq. The comment says "allow through for basic endpoints" but the return happens unconditionally for all gated controllers.
Fix: At minimum, only allow through if the controller is explicitly allow-listed as "open before provisioning" (e.g. a `/me` style endpoint). For all others, deny with 403 when no Shillinq profile exists. Example:
```php
if ($shillinqUser === null) {
    throw new PermissionDeniedException('No Shillinq user profile found');
}
```

---

### Finding 2

**[CRITICAL] Hardcoded `@example.com` placeholder email written to production data**
`lib/Controller/UserController.php:614`
Issue: The provision endpoint constructs `$employeeId.'@example.com'` as the user email. This was clearly a placeholder that was never replaced. When HR provisioning fires in production, every new user will be created with a fake email address, breaking all Nextcloud notification delivery and recertification campaigns.
Fix: Either derive the email from Nextcloud's `IUserManager` lookup of `$employeeId`, or require `email` as a mandatory field in the request body (add to the existing empty-check at line 599).

---

### Finding 3

**[CRITICAL] Frontend violates required CnIndexPage / CnDetailPage architecture**
`src/views/role/RoleIndex.vue`, `src/views/role/RoleDetail.vue` (and all other new `src/views/*` files)
Issue: The design spec (`openspec/changes/access-control-authorisation/design.md`) explicitly mandates the `CnIndexPage` / `CnDetailPage` / `CnFormDialog` pattern from `@conduction/nextcloud-vue`. The implementation uses raw HTML `\<table\>`, `\<form\>`, `\<input\>`, and `\<div\>` wrappers instead. This means consistent UX (search bars, side panels, action menus, pagination, empty-state illustrations) is absent across all 14 new views, and the component contract with the Conduction design system is broken.
Fix: Refactor each index view to use `\<CnIndexPage\>` and each detail view to use `\<CnDetailPage\>` from `@conduction/nextcloud-vue`. Dialogs should use `\<CnFormDialog\>` instead of `\<NcDialog\>` with a raw `\<form\>`.

rubenvdlinde:
```json
{ "pass": false, "blocking": ["Permission cache not keyed by userId — cross-user permission bleed in lib/Service/FieldSecurityService.php:127", "Fail-open on exception in financial enforcement control in lib/Service/PurchasingLimitService.php:130-133"] }
```

rubenvdlinde:
### Finding 4

**[WARNING] Audit log does not capture the acting user**
`lib/Service/AuditLogService.php`
Issue: `AuditLogService::log()` records the action, resource, IP address, and User-Agent, but never records \*who\* performed the action. An audit log without an actor is of limited forensic value. Cross-referencing IP addresses is error-prone in a proxied environment.
Fix: Inject `IUserSession` into `AuditLogService` and add `'userId' => $this->userSession->getUser()?->getUID()` to the `$data` array. The `accessControl` schema's `details` object can carry it in the short term, but ideally add a top-level `userId` field to the schema.

---

### Finding 5

**[WARNING] PurchasingLimitService fails open — exceptions silently allow unlimited purchases**
`lib/Service/PurchasingLimitService.php:104-107`
Issue: The `catch (\Throwable $e)` block returns `true` with the comment "Fail open: allow the purchase if the check fails." If OpenRegister becomes unavailable mid-session, every purchase above the actual limit will be silently approved. For a financial authorisation control this is a business-logic risk, not just a technical one.
Fix: Either fail closed (return `false`) and surface a specific error to the caller, or document the deliberate fail-open policy in the spec and add a monitoring alert. If fail-open is the accepted policy, the controller calling this service must log a WARNING-level event so the failure is observable.

---

### Finding 6

**[WARNING] Invalid / empty dates silently coerce to current time in delegation creation**
`lib/Controller/DelegationController.php:718-719`
Issue: `$startDate` and `$endDate` default to `''`. Passing an empty string to `new \DateTime('')` does not throw; it creates a DateTime representing the current moment. If a client omits these fields, both dates become "now" and the `$end <= $start` guard in `DelegationService::createDelegation()` will block the request — but the error message "End date must be after start date" will be confusing when both fields were simply absent.
Fix: Validate that `startDate` and `endDate` are present and parseable before constructing `DateTime` objects. Return a `422` with a clear field-level error if either is missing or malformed:
```php
if (empty($startDate) || empty($endDate)) {
    return new JSONResponse(['error' => 'startDate and endDate are required'], 422);
}
```

---

### Finding 7

**[WARNING] computeNextRun() heuristic is incorrect for weekly cron expressions**
`lib/BackgroundJob/RecertificationNotificationJob.php:157-177`
Issue: The method classifies a cron expression as "monthly" if `$parts[2] !== '*'` and "daily" if `$parts[1] !== '*'`. A weekly cron such as `0 9 * * 1` (every Monday at 09:00) has `$parts[2] === '*'` and `$parts[1] === '9'` (not `'*'`), so it would be treated as daily — causing the job to dispatch notifications every day instead of once a week. Any cron-aware library (e.g. `dragonmantank/cron-expression`) would resolve this correctly without bespoke parsing.
Fix: Replace the heuristic with a proper cron-expression library. If adding a dependency is out of scope for this PR, at minimum restrict the `cronExpression` schema field to a documented allow-list of supported patterns (`0 9 1 * *`, `0 9 * * 1`, etc.) and document the limitation.

---

### Finding 8

**[WARNING] processReviewDecisions() accesses array keys without isset() guards**
`lib/Service/RecertificationService.php:112-143`
Issue: The method iterates `$decisions` and accesses `$decision['action']` and `$decision['userId']` directly. If a client sends a malformed decision entry (e.g. missing `action` key), this triggers an `Undefined array key` PHP error (or TypeError in strict mode). The controller only validates that `$decisions` is non-empty; it does not validate the shape of individual entries.
Fix: Add per-entry validation before the inner `if`:
```php
if (!isset($decision['action'], $decision['userId'])) {
    continue; // or collect and return validation errors
}
```

---

### Finding 9

**[WARNING] Incomplete test coverage — DelegationService, RecertificationService, and all controllers are untested**
`tests/unit/`
Issue: The four included test files cover `PermissionGateMiddleware`, `FieldSecurityService`, `PurchasingLimitService`, and `DelegationExpiryJob`. Conspicuously absent are:
- `DelegationService` (create/revoke logic, notification dispatch)
- `RecertificationService` (processReviewDecisions, dispatchReviewNotifications)
- `RecertificationNotificationJob` (isDue logic, computeNextRun correctness)
- All 7 controllers (happy path, 4xx paths, auth boundary)

The acceptance criteria in `openspec/changes/access-control-authorisation/tasks.md` require unit tests for each task. Controller tests are especially important for the auth boundary (`@RequiresRoleLevel` annotations) since the middleware only reads these annotations at runtime.
Fix: Add unit tests for the missing services and controllers before merge. Controller tests should mock `ObjectService` and assert HTTP status codes and response shapes.

rubenvdlinde:
```json
{ "pass": false, "blocking": ["CRITICAL: PermissionGateMiddleware fails open for all gated endpoints when no Shillinq user profile exists, bypassing all RBAC checks (lib/Middleware/PermissionGateMiddleware.php:97-101)", "CRITICAL: Hardcoded @example.com placeholder email written to production user records in provision endpoint (lib/Controller/UserController.php:614)", "CRITICAL: All 14 new frontend views use raw HTML tables/forms instead of mandated CnIndexPage/CnDetailPage/CnFormDialog components from @conduction/nextcloud-vue", "WARNING: AuditLogService does not record the acting user identity, rendering audit log entries non-attributable (lib/Service/AuditLogService.php)", "WARNING: PurchasingLimitService fails open on any exception, silently allowing unlimited purchases when OpenRegister is unavailable (lib/Service/PurchasingLimitService.php:104-107)", "WARNING: Empty or missing startDate/endDate in delegation creation silently coerce to current time instead of returning a 422 (lib/Controller/DelegationController.php:718-719)", "WARNING: computeNextRun() heuristic misclassifies weekly cron expressions as daily, causing over-frequency notification dispatch (lib/BackgroundJob/RecertificationNotificationJob.php:157-177)", "WARNING: processReviewDecisions() accesses decision array keys without isset() guards, risking undefined-key errors on malformed input (lib/Service/RecertificationService.php:112-143)", "WARNING: Unit tests missing for DelegationService, RecertificationService, RecertificationNotificationJob, and all 7 controllers"] }
```

rubenvdlinde:
## Hydra Security Review — Clyde Barcode

**Semgrep SAST** (p/security-audit, p/secrets, p/owasp-top-ten): **0 automated findings** across both changed files.

---

### Manual OWASP Review

---

**[CRITICAL] Permission cache not keyed by userId — cross-user permission bleed**
Rule: OWASP A01:2025 – Broken Access Control / ASVS V4.1.1
`lib/Service/FieldSecurityService.php:127`
Issue: `$permissionCache` is a single instance-level field. `loadPermissions()` checks only `if ($this->permissionCache !== null)` before returning the cache, ignoring the `$userId` argument on subsequent calls. If the service instance is reused across multiple users in one PHP lifecycle (background jobs, CLI batch processing, admin bulk operations, or any scenario where one request processes multiple users' data), user A's permission matrix is silently applied to user B. This causes both unauthorised data exposure (fields visible that should be hidden) and unauthorised write escalation (fields writable that should be blocked).
Fix: Key the cache by userId: change `private ?array $permissionCache = null;` to `private array $permissionCache = [];` and replace the cache guard with `if (isset($this->permissionCache[$userId])) { return $this->permissionCache[$userId]; }`, storing results as `$this->permissionCache[$userId] = $allPermissions;`.

---

**[CRITICAL] Fail-open on exception in FieldSecurityService — all fields exposed and all writes permitted on error**
Rule: OWASP A10:2025 – Mishandling of Exceptional Conditions / ASVS V1.1 (Fail Secure)
`lib/Service/FieldSecurityService.php:173`
Issue: The `catch (\Throwable $e)` block sets `$this->permissionCache = []` and returns an empty array. `filterResponse()` receiving an empty permissions array iterates zero times and returns the **original unfiltered object** — every field is exposed. `checkWritePermission()` receiving an empty permissions array finds no matching deny-rule and returns `true` — every write is permitted. If `ObjectService` is unavailable (e.g. database down, service misconfiguration), the entire field-level access control layer is silently bypassed.
Fix: Fail closed. On exception, either throw (let the caller handle it with an HTTP 500) or return an explicit "deny-all" sentinel. For `filterResponse()` this means returning `[]` (empty object) on error. For `checkWritePermission()` this means returning `false` on error. Log the exception before returning.

---

**[CRITICAL] Fail-open on exception in PurchasingLimitService — financial controls bypassed on error**
Rule: OWASP A10:2025 – Mishandling of Exceptional Conditions / OWASP A06:2025 – Insecure Design
`lib/Service/PurchasingLimitService.php:130`
Issue: The `catch (\Throwable $e)` block returns `true`, explicitly allowing the purchase when the limit-check itself fails. The inline comment acknowledges this: *"Fail open: allow the purchase if the check fails."* Any transient or persistent failure in `ObjectService` (database timeout, container misconfiguration, malformed role data) will silently grant unlimited purchasing authority to every user. This is a financial business-logic control — fail-open here means the safeguard provides zero protection under adversarial or fault conditions.
Fix: Fail closed. On exception, return `false` (block the purchase) and surface an appropriate error to the caller so it can respond with a retryable error rather than silently approving. If availability is a concern, implement a circuit-breaker or retry pattern upstream rather than fail-open here.

---

**[WARNING] No roles resolves to unrestricted purchasing**
Rule: OWASP A06:2025 – Insecure Design / ASVS V11 (Business Logic)
`lib/Service/PurchasingLimitService.php:85`
Issue: When `$roleIds` is empty (user has no active access rights), `checkLimit()` returns `true` unconditionally (line 86–87). A user with zero delegations or whose access rights have all lapsed can make purchases of any amount without restriction. This is likely unintentional — a user with no active role should arguably be the most restricted, not the least.
Fix: Return `false` (block all purchases) when `$roleIds` is empty, unless the business requirement explicitly states that role-less users are unrestricted (in which case document it clearly and consider whether that is the intended design).

rubenvdlinde:
```json
{ "pass": false, "blocking": ["Permission cache not keyed by userId — cross-user permission bleed in FieldSecurityService.php:127", "Fail-open on exception in FieldSecurityService — fields exposed and writes permitted on error at FieldSecurityService.php:173", "Fail-open on exception in PurchasingLimitService — financial controls bypassed on error at PurchasingLimitService.php:130"] }
```

PR #495, opened 2026-08-10
refactor(commitment)!: Verplichting demanded two vocabularies at once — fix it, then move the whole domain to English

rubenvdlinde:
## 1. The defect that started this

Two register fragments both declared `Verplichting.required`. ADR-037's `deepMergeConfig()` merges by schema key and **concatenates** list values, so the stored schema demanded *both* naming conventions — the duplicated entries are the tell:

```
required = [verplichtingNummer, omschrijving, bron, bedrag, administrationId, status,
            administrationId, verplichtingsnummer, soort, status]
```

No payload of either model could satisfy it, so **both sides failed, each blaming the other's fields**: seeds skipped on `occ app:enable`, and Newman's `REQ-001` POST returned 400.

It was not only test data. `TenderNedAwardDetectedListener` wrote the wrong spelling and no `soort`, so **every award produced a rejected object** — and its `catch` is fail-soft, so TenderNed award detection **silently created nothing**.

**Measured with the repo's own checker: 74 → 70 unimportable seeds.** `BASELINE` lowered to 70.

## 2. Then: the whole domain to English

Per Ruben — schemas are English, Dutch lives in the **translations**. There is no standardised-term exemption for internal code and schemas; only statutory *wire* formats stay Dutch, at the adapter boundary.

`Commitment` is not invented: `CommitmentMaterialisationService`, `CommitmentMaterialisationListener` and `BudgetLineCommitments.vue` already used it.

| | |
|---|---|
| Schemas | `Verplichting`→`Commitment`, `Verplichtingsregel`→`CommitmentLine`, `Verplichtingsmutatie`→`CommitmentMovement`, `Mandaat`→`Mandate`, `Goedkeuringsstap`→`ApprovalStep`, `TenderNedAanbesteding`→`TenderNedProcurement`, `OpdrachtUitvoering`→`OrderFulfilment` |
| Classes | 6 lifecycle guards/listeners + their test classes |
| Also | every property, enum value, lifecycle transition, seed slug, display title, the register declaration, the UI manifest, and the full test suite |
| Translations | 42 titles added to `l10n/nl.json` — the Dutch that was hard-coded in the schema now lives where it belongs |

Full map: `docs/commitment-domain-english-rename-map.md`.

## 3. Things that would have broken silently

**A shared field contract.** `purchase-requisition.json` documents that `Requisition` carries Commitment's exact field names so `BudgetBlocker::canCommit` runs unmodified against both. Renaming only Commitment wouldn't have errored — the guard reads with `??`, so every requisition would have been checked against a **zero budget figure**. Both sides renamed.

**Guards are wired by class name from the fragments.** A lifecycle guard that cannot be resolved doesn't fail loudly; it stops guarding.

**A schema-name string comparison.** `CommitmentTransitionListener` lower-cases the incoming schema and compares it to a literal. Left unrenamed it would simply never fire again.

**A formula over renamed fields.** Budget's `vrije_ruimte` still read `geautoriseerd_bedrag - gerealiseerd_bedrag - openstaande_verplichtingen`.

**Named arguments defeat `php -l`.** PHP resolves them at runtime, so the file reports clean and fatals only when that branch executes. Found several by diffing call-site arg names against parameter lists.

## 4. The wire boundary is deliberate

`$payload[...]` in the TenderNed listener keeps **Dutch** keys — `aanbestedingId`, `titel`, `contractWaarde`, `status='gegund'` — because that is the event shape TenderNed sends. The array literal there has English keys and Dutch lookups on purpose, with a comment. `sample-tenderned-aanbiedingen.json` stays Dutch for the same reason.

Mid-change my own bare-word replace rewrote `$payload['aanbestedingId']` → `procurementId`, detaching the listener from the only field identifying the tender. Caught by diffing payload reads against the fixture keys; both now agree on all 8.

## 5. Verification

```
validate-seeds     PASS — 70 (baseline 70, was 74)
validate-registers PASS — 493 schemas, no slug collisions
check-l10n         OK   — was FAIL with 2 orphaned keys
validate-manifest  PASS
php -l             every changed file
```

⚠️ **Not verified locally:** PHPUnit (the dev container bind-mounts the main checkout, so it does not see this worktree), vitest (`@vitejs/plugin-vue` absent), and `validate-manifest`'s Ajv pass (ajv absent — it fell back to structural lint and said so). **CI is the authority on those three.**

⚠️ #484 also edits `BASELINE`. Whichever merges second must take the **lower** number.

## 6. Scope

Commitment/TenderNed domain only. The other ~23 Dutch-named schemas (Voorziening, Reserve, Taakveld, Uren*, Innovatiebox, ENSIA, rechtmatigheid …) are separate slices; `ontmantelingsVerplichting` is Voorziening's and deliberately untouched. ~43 pre-existing dangling `@spec` targets were left alone — only the 7 in scope were repaired.

## 7. The data migration this PR was missing

Everything above renames the *declaration*. Nothing migrated the *data* — and a property rename **is** a data migration.

OpenRegister reads and writes only what the schema declares. The moment `Verplichting.verplichtingsnummer` stops being declared, every stored value under it goes invisible: no error, no warning, CI fully green, and the number is simply gone from every screen. The **schema** rename is worse — the importer resolves schemas by slug (`ImportHandler::findByApplicationAndSlug`), so a changed slug does not rename the schema, it **creates a new empty one** and strands every existing `Verplichting` / `Verplichtingsregel` / `Mandaat` / … object under a slug the app never asks for again.

`lib/Repair/MigrateCommitmentDomainToEnglish.php` is the migration, registered in `appinfo/info.xml` immediately after `InitializeSettings` (which is what creates the English schemas). **Two passes off one declarative map:**

1. **Slug move** — stranded objects are re-pointed from the old slug to the new one **keeping their uuid**. Not cosmetic: the whole domain joins on it (`CommitmentLine.commitment`, `CommitmentMovement.commitment`, `ApprovalStep.commitment`, `TenderNedProcurement.commitmentId`, `OrderFulfilment.commitmentId`), so a copy under a fresh uuid would orphan every child row.
2. **In-place key move** — objects already under the new slug but carrying legacy keys. This also covers **`Budget` and `Requisition`, whose slugs never changed but whose properties did**. That matters: `BudgetBlocker` reads `authorisedAmount` / `realisedAmount` / `outstandingCommitments` with `??`, so a half-migrated Budget silently checks every commitment against a **zero ceiling**.

**Version bumped 0.9.23 → 0.9.24, and that is load-bearing.** Nextcloud only runs `post-migration` repair steps when the installed version differs. Shipping this under an unchanged 0.9.23 would register a data migration that never runs on any instance already at 0.9.23.

### The split brain, resolved

`verplichtingsnummer` (declared by the owning fragment, read by `BudgetBlocker`) and `verplichtingNummer` (written by `TenderNedAwardDetectedListener`) are **separate columns**, so real data may carry either — or both, disagreeing. Same for `looptijd_van`/`looptijdStart` and `looptijd_tot`/`looptijdEind`.

**Rule:** sources are ordered per target property and the **first non-empty wins, declared spelling first** — that is the value every reader in the app actually consumed. A genuine disagreement applies the winner **and warns with the object uuid and both values**. The loser is not discarded: this step never clears a legacy column, and OpenRegister writes only *declared* columns on update, so the losing value stays exactly where it was and remains recoverable.

Also handled: **`InitializeSettings` runs first and reseeds the mandate templates** against a freshly-created (empty) `Mandate` schema, creating every one. Migrating the instance's legacy `Mandaat` rows on top would leave **two mandates per code** — not cosmetic for a signing-authority lookup. A `NATURAL_KEYS` gate (`administrationId` + `mandateCode`, and equivalents per schema) recognises those and leaves the legacy row in place instead of duplicating it. An *incomplete* key is abandoned rather than padded, so a row with no business identity still migrates.

### Proven both directions, per property

`tests/Unit/Repair/MigrateCommitmentDomainToEnglishTest.php` — **187 tests, 879 assertions.** For each of the **100+ (schema, English property, Dutch source key) triples** and each of the **54 enum value renames**, generated from the map rather than hand-picked:

> seed the value under the **Dutch** name → assert it is **NOT readable** under the English name → run the step → assert it **is**.

The "before" leg is the point. A migration test that only checks the post-state cannot tell a working migration from a fixture that was already migrated.

**The suite was shown able to fail before it was trusted.** Eight planted defects, each confirmed red, then green again on revert:

| planted defect | caught by |
|---|---|
| declared spelling dropped from `commitmentNumber` sources | 6 failures |
| split-brain precedence reversed (last wins) | 1 |
| never-clobber guard removed | 2 |
| enum value pass skipped | 54 |
| slug-move idempotency gate dropped | 1 |
| `0` / `false` treated as absent | 1 |
| natural-key dedup disabled | 1 |
| incomplete natural key padded instead of abandoned | 1 |

### Idempotent three ways, non-destructive

Already migrated (target values never overwritten; a uuid *or* business identity already under the target is skipped whole) · never written under either name (**no save at all** — an already-migrated object still reads its retained legacy column back on every run, so counting that as a change would re-save the whole domain forever) · written under both (precedence, once).

Nothing is deleted, no legacy column is cleared, **no `required` list is relaxed**. An object that cannot satisfy the target schema — a legacy commitment that never carried `soort`, say — fails its own save, is reported by uuid, and is left for an operator. It is not forced through by weakening the schema.

### Also fixed here

`src/manifest.d/20-tenderned-integratie.json` renamed the schema and five of its six column keys but kept **`verplichtingNummer`** in the Commitments column list (twice) and in `defaultSort`. That page would have rendered a blank Number column and sorted on a field that no longer exists. Rename-only, in a file this PR already owns.

### Verification

```
phpunit (unit)     4214 tests, 43671 assertions, 0 failures   (was 4027)
  new suite          187 tests,   879 assertions, 8/8 planted defects caught
phpcs / phpmd / psalm / phpstan   clean on both new files; psalm + phpstan: no errors repo-wide
check:registers    PASS
check:seeds        PASS — 70 (baseline 70) — unchanged before and after
check:manifest     PASS — Ajv 0 errors, consistency 0 issues (ajv IS present now)
hydra gates        65 gate lines, 10 FAIL — byte-identical before and after this change
named-arg audit    clean — no call site passes a renamed-away parameter to a renamed method,
                   and no call to a removed method name survives
```

### 8. The UI still pointed at the five schemas this rename deletes — now fixed and measured

`src/manifest.d/bookkeeping-verplichtingenadministratie.json` was never touched by this PR and still named `Verplichting`, `Verplichtingsregel`, `Verplichtingsmutatie`, `Mandaat`, `Goedkeuringsstap` — schemas that stop existing the moment this rename lands. Nine pages and related-lists pointed at nothing. `src/manifest.json` carried four Dutch `Requisition` keys the register had already renamed.

**Why a green gate report did not catch it.** gate-53 runs its cross-reference stage **only when its structural stage passes**. This repo's structural stage sits at 120, so the crossref stage never ran, and the gate line read identically before and after. Running the two stages directly:

| gate-53 stage | before | after |
|---|---|---|
| structural | 120 | **120** (unchanged — still masking until #503 lands) |
| crossref | 161 | **152** (−9) |

The nine cleared findings are precisely the five renamed schemas; **nothing was introduced**. Eleven undeclared-schema findings remain, all pre-existing and unrelated (`Kostenpost`, `VATAuditRecord`, `VATGLAccounts`, `WBSO*`).

**Exactly what changed:** schema slugs, property keys, and the `filterField`s that join them. In `src/manifest.json`, exactly seven key lines: `programma`→`programme` (×2), `boekjaar`→`fiscalYear` (×2), `soort`→`commitmentType` (×1), `totaalbedrag_excl_btw`→`totalAmountExclVat` (×2) — all on `Requisitions` / `RequisitionDetail`.

**Deliberately NOT changed:** page ids, routes, `detailRoute`/`indexRoute`, menu entries. The rename does not invalidate them and they are cross-referenced from `src/menu-layout.json` and `src/manifest.json`'s page list, which another agent is editing.

Also regenerates **`src/manifest.d.shell.json`**, which this PR left stale — it renamed page ids in `20-tenderned-integratie.json` without re-running `scripts/generate-manifest-shell.js`, so the checked-in boot-eager projection `src/main.js` consumes still advertised `TenderNedAanbestedingen` and `Verplichtingen`. Preventing that silent drift is the generator's stated reason to exist. Note gate-53's crossref **does** cross-reference schema slugs; what nothing checks is **column/property keys**, which is why the `manifest.json` half was invisible to every gate.

### ⚠️ Known-remaining (not fixed here)

**Still Dutch in the schema, so deliberately still Dutch in the manifest** — renaming these in the UI would break working columns to make a file look consistent. Each needs its own register rename *and* a matching entry in the repair step's map when it moves:

- `Budget.vrije_ruimte` — and nothing computes it; there is no calculation block, despite the description naming the new English inputs.
- `Mandate`: `houder`, `houder_type`, `uitsluitingen`, `geldig_van`, `geldig_tot`, `vastgesteld_bij`
- `ApprovalStep`: `rol_vereist`, `toegewezen_aan`, `behandeld_op`, `opmerking`, `vereist_handtekening`, `handtekening_bestand`

Dutch keys in *other* slices (`bookkeeping-rechtmatigheidsverantwoording.json`, `bookkeeping-titel-9-jaarrekening.json`) belong to schemas outside this domain and are untouched.

### 🔴 BLOCKING before merge — two mechanisms observed only in source, never run

**Not verified live, and I could not verify them here.** The migration is unit-proven against a fake ObjectService. Two mechanisms are read from OpenRegister's source rather than observed, and both decide whether this migration preserves or destroys data:

1. **`saveObject(schema: <new>)` on an existing uuid re-points the object rather than copying it.** Read from `SaveObject::handleObjectUpdate`, which sets register+schema from the passed context. **If this is wrong, every migrated object gets a fresh uuid and every child row is silently orphaned** — `CommitmentLine.commitment`, `CommitmentMovement.commitment`, `ApprovalStep.commitment`, `TenderNedProcurement.commitmentId`, `OrderFulfilment.commitmentId` all join on it.
2. **Obsolete columns survive `syncTableForRegisterSchema`.** Read from MagicMapper: sync only ADDS columns and makes obsolete ones nullable, and the read path is `SELECT *`. **If this is wrong, the old column is dropped before the migration can read it** — and the split-brain "loser value stays recoverable" guarantee is void with it.

**The test that must be run** (a shape check is not enough — a copy produces a plausible-looking row too, so the **uuid** is the discriminator):

- seed a `Verplichting` with at least one `Verplichtingsregel` pointing at it, under the OLD schema;
- run the migration;
- assert the **same uuid** now resolves under `Commitment`;
- assert **the children still resolve to that parent** — this is what tests re-pointing rather than copying;
- separately, assert the obsolete column is still present **with its data** after the schema sync.

I could not stand an instance up: **the host disk is at 100%** (24 MB free at its worst; ~400 MB after I released my own build artifacts), with thirteen containers belonging to other agents and the shared dev instance live on it. The only prunable image is untagged but is **the running image of two other agents' containers**. Standing up a NC+Postgres stack plus a two-version upgrade rehearsal on that disk risks filling it completely and taking the shared environment and other agents' work down with it. This needs a host with room; it is not a claim I am willing to soften to reach a merge.
Refs #485, #471


rubenvdlinde:
## Not merged in the fleet consolidation — this one would lose data

Every other open PR touched in the last week was merged into `development` today (33 of them, several with `--admin` while red). This one and pipelinq#779 were deliberately held back, because the failure here is not red CI — it is silent data loss plus a language regression.

### What changed underneath this branch

`development` has since merged its own Dutch→English programme for exactly these schemas, each with a migration:

- #533 — the `bedrag` cluster → `amount`
- #534 — 137 more property names
- #541 — 736 names from a token dictionary
- #552 — 135 pre-existing Dutch names
- #554 — the 25 tranche 2 could not reach
- #560 — enum values, Dutch kept in l10n

So this is no longer *the* rename. It is a **second, competing rename of already-renamed-and-already-migrated properties.**

### 1. The migration maps from keys that no longer exist

`MigrateCommitmentDomainToEnglish` maps from the **original** Dutch keys. It has no entry for any of the intermediate keys `development` migrated the data to:

| key on `development` today | in this branch's migration map |
|---|---|
| `term_from`, `term_to` | ✗ |
| `commencementDate` | ✗ |
| `vat_number` | ✗ |
| `total_amount_excl_vat`, `total_amount_incl_vat` | ✗ |
| `kind` | ✗ |

On any install that has run `development`'s migrations — which is every install, once `development` ships — the stored objects live under those keys. This branch's schemas then declare `termStart`, `commitmentDate`, `vatNumber`, `totalAmountExclVat`, `commitmentType`, and the migration goes looking for Dutch keys that are already gone. It finds nothing, reports success, and the values become invisible.

That is precisely the failure this branch's own commit 30b391f5 describes: *"OpenRegister reads and writes only what the schema declares, so the moment `Verplichting.verplichtingsnummer` stops being declared, every stored value under it is invisible — silently, on every existing install, with CI green."* The reasoning is right; it was written against a `development` that has since moved.

### 2. It reverts names to Dutch

`development` translated these; this branch would put them back:

| schema | `development` | this branch |
|---|---|---|
| Mandate | `holder`, `holder_type` | `houder`, `houder_type` |
| Mandate | `valid_from`, `valid_to` | `geldig_van`, `geldig_tot` |
| Mandate | `determined_at` | `vastgesteld_bij` |
| ApprovalStep | `remark`, `role_required` | `opmerking`, `rol_vereist` |
| ApprovalStep | `handled_on`, `assigned_in` | `behandeld_op`, `toegewezen_aan` |
| ApprovalStep | `required_signature`, `signature_file` | `vereist_handtekening`, `handtekening_bestand` |
| Budget | (free space) | `vrije_ruimte` |

Against the standing rule that all code is English, that is a regression, not a merge conflict.

### What this needs

Not a conflict resolution — a rebase onto the vocabulary `development` now has. Concretely:

1. Re-derive the target names from `development`'s current schemas rather than from the pre-#533 Dutch ones, and keep English where `development` already has it.
2. Extend `MigrateCommitmentDomainToEnglish` with the intermediate→final pairs above, so installs migrated by #533–#560 are carried forward rather than stranded.
3. Reconcile with `RenameDutchColumnsMap`/`RenameDutchColumns` so the two repair steps have a defined order and do not each half-move the same object.
4. Keep the remaining genuinely-new work here (the gate-54 relation-block retirement, the gate-57 suppressions, the `@spec` fixes) — none of that is affected.

`development` is currently *half*-translated — schema names are still Dutch (`Verplichting`, `Verplichtingsregel`, `Mandaat`, `Goedkeuringsstap`) and properties are a mix of camelCase, snake_case and Dutch. Finishing that is worth doing. This branch is the right intent measured against the wrong baseline.

rubenvdlinde:
**Measured, so the size is on the record rather than implied:**

| | |
|---|---|
| commits on `development` since this branched | **239** |
| conflicted files | **51** of 62 |
| conflict hunks | **363** |
| `lib/Settings/register.d/*.json` conflicted | **3** |

## Why I did not resolve it

I attempted the merge and stopped. Taking either side mechanically produces a wrong result, in both directions:

- **`-X ours` (this branch)** drops 239 commits of development's own work in these files — including `LeaseActivationListener` and the `OpdrachtUitvoering*` rename that landed independently.
- **`-X theirs` (development)** undoes the rename that is this PR's entire purpose.

The conflicts are not incidental. Both sides renamed the *same* vocabulary in *different* directions — this branch to `CommitmentLine` / `programme` / `financial_year` / `TenderNedProcurementGuard`, development to `Verplichtingsregel` / `programma` / `boekjaar` / `TenderNedAanbestedingGuard`. Reconciling that is a judgement call about which names win, not a merge.

And three of the conflicts are in `register.d/` schema files. A property rename there is a **data migration**, not a text edit: existing objects carry the old property names, and nothing in this PR migrates them.

## Recommendation: rebuild, don't merge

A vocabulary rename is cheap to re-apply and expensive to reconcile. Re-running it against current `development` gives a clean, reviewable diff; merging 363 hunks gives a diff nobody can review and a schema change nobody can verify.

Concretely:
1. Branch fresh from `development`.
2. Re-apply the Dutch→English rename (the standing "all code is English" rule) to the code as it now stands.
3. Ship the `register.d` renames **with** a migration for existing objects, or split them into their own PR so the data change gets its own review.

Happy to be overruled — if this should be force-merged as-is, say so and I will resolve it hunk by hunk. But it should be a deliberate decision, not the by-product of an auto-merge sweep.

_Assessed during a fleet merge session; no changes pushed to this branch._

rubenvdlinde:
**Status check — this PR reads greener than it is.**

The checks panel shows three green ticks and nothing red:

```
COMPLETED/SUCCESS  Analyze (actions)
COMPLETED/SUCCESS  Analyze (javascript-typescript)
COMPLETED/SUCCESS  CodeQL
```

All three are CodeQL. `Code Quality` — phpcs, phpstan, psalm, phpmd, PHPUnit, Hydra Gates, Newman, Playwright — has produced **no verdict at all** on this branch. Nothing has failed because nothing has been measured, and an absent gate looks exactly like a passing one in that list.

Where the branch actually sits:

```
$ gh api repos/ConductionNL/shillinq/compare/development...fix/fragment-vocabulary-group-a
  ahead_by=30  behind_by=462  status=diverged
  mergeable: CONFLICTING
```

462 commits behind, opened 2026-08-10, last touched 2026-08-20.

That matters more than usual for this change specifically. It is a `!` breaking refactor of a domain vocabulary (Verplichting), and `development` has moved a long way underneath it — including, in the last day alone, the PHPStan 2 migration across the whole app, the budget-scenarios and budget-known-costs work, the `check:manifest` dead-config-key fix (#1040), the gate-69 page-type fix (#1051), and the RECURRING_* expander correction (#1037). Several of those touch the same service layer.

**Deliberately not resolved here.** Re-baselining a diverged breaking refactor across 462 commits is not a conflict resolution; in practice it is a re-implementation against a different codebase, and the result would need the full gate set plus a human read of the vocabulary decision before it could be trusted. Doing that unsupervised and merging it on `--admin` is the wrong shape of risk for a `!` change.

What it needs, concretely:

1. A decision on whether the two-vocabulary split is still the intended design after the budget work landed.
2. A fresh branch off current `development` carrying the 30 commits' intent, rather than a merge of the existing one.
3. A real `Code Quality` run — the branch has never had one, so the 30 commits are unverified against any gate.

Flagging rather than touching it, so the green ticks are not mistaken for coverage.

PR #556, opened 2026-08-15
refactor(deps): type-hint OpenRegister's published contract (ADR-084)

rubenvdlinde:
## What

Services take `OCA\OpenRegister\Contract\ObjectServiceInterface` instead of the
concrete class, bound in the composition root, with `conduction/hydra-gates`
bumped to **v1.8.0** so composer installs the interface into `vendor/`.

## Why this is what unblocks ADR-083

A leaf app cannot load a class from another Nextcloud app, so a typed
constructor parameter had no satisfiable double:

```
TypeError: __construct(): Argument #N ($objectService) must be of type
OCA\OpenRegister\Service\ObjectService, class@anonymous given
```

All seven ADR-083 rollout PRs stopped there. Measured across the fleet, **ten of
sixteen consumers had hand-rolled a double** rather than solve it — 0–13 methods
each against a real class of **88**.

`ObjectServiceInterface` is 25 methods, scoped by measuring per CLASS rather than
per call: a class can only type-hint the interface if *every* method it calls is
on it. Per call, eight methods look sufficient; per class they cover 58% and
strand 414 consumer classes.

## What is deliberately NOT touched

Files naming `'OCA\OpenRegister\Service\ObjectService'` only as a **container
key**. Those are availability-guarded lookups — the ADR-083 rule-1 exception —
and the string must go on naming the concrete service, because that is the key
the alias resolves *to*. Rewriting it would make the container look up something
nobody registered.

## The binding

`registerServiceAlias`, not a factory: it resolves when something asks for the
interface, so an instance without OpenRegister fails at the route that needed the
data rather than at registration. Both names are strings and neither triggers an
autoload — which is what keeps ADR-083 rule 3's promise that the start screen
still boots.

## On the transformer

The first version masked string literals with a **regex**, and it silently
under-converted: in one file the comment `// King's Day.` opened a string that
did not close until `date('N')` 240 lines later, so every type position between
them looked quoted and the file was skipped. **A skipped file is
indistinguishable from a clean one** — it took comparing counts to notice.

It now takes comment and string ranges from PHP's own `token_get_all()`, lints
every file with `php -l` and reverts on failure, and reports every file it
declines to convert together with the reason.

## Related

- ADR-084 — ConductionNL/hydra#576 (merged)
- interfaces + `implements` — ConductionNL/openregister#2498 (merged)
- gate-67 parity + shipped copy — ConductionNL/.github#463 (merged), v1.8.0
- first consumer, proven green — ConductionNL/docudesk#462 (merged)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rubenvdlinde:
Correction to commit `3d3b73ad`: its message describes removing a stale
`container:` argument, but the change it carries is the opposite — one test
construction gaining the `objectService` argument. Zero containers were removed
in that commit (the 134 were removed earlier, in `623d91b5`).

The site only became visible after the container removal reshaped the
surrounding call, which is why it appears in a second pass. The tree is now 0
for both problems:

```
missing required argument : 0
stale container argument  : 0
classes legitimately still taking a ContainerInterface : 151
```

PR #559, opened 2026-08-15
feat(shillinq): translate enum values to English, with the Dutch kept in l10n

rubenvdlinde:
Translates shillinq's enum **values** to English (batch 3), keeps the Dutch word visible to Dutch users via l10n, and migrates the rows already stored.

## Why this one took four attempts

Three earlier runs produced 27, 58 and 21 regressions and were reverted. The applier was never the problem — the **map** was:

| defect | count | what it did |
|---|---|---|
| map key named a property no schema declares | **31 of 90** | `frequency` matched only a local variable, so `CashflowRecurringGuard` moved to `ANNUALLY` while its enum, seeds and fixtures stayed `JAARLIJKS` — the annual branch silently stopped firing |
| entry only recased an already-English token | **56** | `ACMReport → aCMReport`, `WEEK → wEEK` — renaming entity types and schema keys |
| replacement changed the casing convention | **168** | `BEHAALD → achieved` in an UPPER_SNAKE enum is a second contract change riding along |

All three are fixed in the map; the 30 ghosts were re-keyed onto their real declaring properties.

## Applier gap this exposed

`default`, `const` and `example` carry the same vocabulary as the enum beside them but live in the property block, so the enum pass never saw them. **16 defaults** were left naming a member their own enum no longer declared — valid JSON, no lint finding, and OpenRegister stamps the stale value onto every new object. Development has 0 such defaults; this branch now has 0.

## Data migration

`RenameDutchColumns` moves the columns; the new `RenameDutchValues` moves what is in them. Once the schema says `ACTIVE`, a row holding `ACTIEF` does not error — a filter on the new value returns null and the caller reads that as "nothing to do".

Logic sits in `RenameDutchValueDecisions` and storage behind `ValueMigrationPort`, because `createMock(IDBConnection)` cannot be built here: PHPUnit resolves the interface's parameter types while *constructing* the double and `Doctrine\\DBAL\\ParameterType` is absent.

## Naming call

`uitbetaald` → **`disbursed`**, not `paid`: `Paid` was already taken in `nl.json` by `Betaald`, and reusing it would show a Dutch reader the wrong word for a state meaning the money went out.

## Verification

- PHPUnit **OK (4122 tests, 43841 assertions)**; identical test *and* assertion count to this branch's own baseline before the pass, so nothing stopped running
- PHPCS **0 errors** on every touched file
- orphaned-default sweep: **0** across 158 register files (development: 0)
- 1394 substitutions across 172 files, every one property-anchored


rubenvdlinde:
Superseded by #560.

This branch forked on 08-12. Development has since taken three property-**name** waves (#541, #552, #554) plus the eslint-10 migration, so merging produced **445 conflicted files** — and `--admin` does not override a conflict.

The deeper problem is that a value map is keyed on property *names*, and every Dutch property this one relied on (`frequentie`, `ernst`, `urgentie`, `huidigStap`, `deelnemerType`) now has zero declarations on development. PHPUnit was green the whole time — against a three-day-old tree.

#560 replays the same change onto current development, re-keyed against the properties that actually exist, and additionally fixes three applier gaps this exposed (orphaned defaults, stale descriptions, and the applier rewriting the migration's own map).

PR #564, opened 2026-08-15
ci: run the 53 Playwright specs this repo has never executed

rubenvdlinde:
`enable-playwright` defaults to **FALSE** in the shared workflow and was never set here, so **"E2E Tests (Playwright)" has reported `skipped` on every run shillinq has ever produced** — while the tree ships a root `playwright.config.ts` and **53 spec files** under `tests/e2e/`.

A skipped job is not a pass. It is the absence of a measurement, and the Quality Report renders it identically to a green one. This is the same shape as `enable-hydra-gates`, two comments further down in the same file, which was fixed for exactly this reason.

I found it while checking that the e2e evidence behind the vocabulary PRs was real: #560's Playwright check reads `skipping`, and the job log is 215 bytes.

## Why a second config, not just the flag

The shared workflow runs `npx playwright test --config=$CONFIG` with **no `--project`**, so whichever config it picks, *every* project in it runs. The root config declares three, and two cannot pass on a runner:

| project | |
|---|---|
| `chromium` | the regression suite — the one CI wants |
| `docs-capture` | re-shoots every journeydoc screenshot; has its own dedicated job |
| `visual` | pixel baselines, host-font/GPU specific, will not byte-match a CI runner |

`playwright-test-path: tests/e2e` makes the workflow's first config lookup hit the new file, which declares only the regression project. The root config is untouched and stays the entry point for local runs and `npm run test:e2e:docs`.

Report paths point at the **app root** deliberately: the workflow uploads `<app>/playwright-report/`, so with the root config's relative paths the upload step matches nothing and silently produces an empty artifact — a failing run with no report to read.

## ⚠️ Expect red on the first run

These 53 specs have never executed in CI. Whatever they report is a **finding, not a regression from this commit**. Two settings are chosen so that first run is informative:

- `retries: 0` — a retry only ever converts red to green; the first run should show what is genuinely broken rather than what is merely flaky.
- `globalTimeout: 38m` inside the job's 45-minute cap — a job the runner kills reports **cancelled**, which is no verdict at all. Playwright stops first and exits non-zero with a tally.

The timeouts start at decidesk's proven values because there is no measured slowest-pass for this suite yet; they should be re-derived from the first green run rather than left as a guess.

I am **not** merging this one without you: unlike the other PRs tonight it can only make the pipeline redder, and the right response to its output is triage, not a rubber stamp.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -113,6 +113,26 @@ jobs:
       enable-frontend: true
       enable-eslint: true
       enable-phpunit: true
+
+      # ── E2E browser tests ────────────────────────────────────────────────
+      # `enable-playwright` defaults to FALSE and was never set here, so the
+      # "E2E Tests (Playwright)" job has reported `skipped` on every run this
+      # repo has ever produced — while the tree ships a root
+      # `playwright.config.ts` and 53 spec files under `tests/e2e/`. A skipped
+      # job is not a pass: it is the absence of a measurement, and the Quality
+      # Report renders it identically to a green one.
+      #
+      # `playwright-test-path` does double duty in the shared workflow: it is
+      # both the test directory AND the first place it looks for a config
+      # (`${playwright-test-path}/playwright.config.ts`, falling back to the
+      # repo root). We ship tests/e2e/playwright.config.ts precisely so that
+      # lookup hits it — the workflow passes no `--project`, so if the ROOT
+      # config were picked it would also run `docs-capture` (re-shooting every
+      # journeydoc screenshot, which has its own job) and `visual` (pixel
+      # baselines that cannot byte-match a CI runner). See that file's header.
+      enable-playwright: true
+      playwright-test-path: tests/e2e
+
       enable-newman: true
       # Shillinq delegates ALL object CRUD to OpenRegister (@conduction/nextcloud-vue
       # store factories hit /apps/openregister/api/objects), so a Nextcloud without
--- a/tests/e2e/playwright.config.ts
+++ b/tests/e2e/playwright.config.ts
@@ -1,0 +1,130 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Shillinq Contributors
+ * SPDX-License-Identifier: EUPL-1.2
+ *
+ * CI regression config for the shared `E2E Tests (Playwright)` job.
+ *
+ * WHY THIS FILE EXISTS
+ * --------------------
+ * `enable-playwright` defaults to FALSE in the shared workflow and was never
+ * set for this repo, so the "E2E Tests (Playwright)" job has reported
+ * **skipped** on every run shillinq has ever produced — while the tree ships a
+ * root `playwright.config.ts` and 53 spec files under `tests/e2e/`. A skipped
+ * job is not a pass; it is the absence of a measurement, and it looks exactly
+ * like a green one in the PR rollup.
+ *
+ * WHY A SECOND CONFIG RATHER THAN JUST FLIPPING THE FLAG
+ * -----------------------------------------------------
+ * The shared workflow runs the suite as:
+ *
+ *     CONFIG="${{ inputs.playwright-test-path }}/playwright.config.ts"
+ *     if [ ! -f "$CONFIG" ] && [ -f "playwright.config.ts" ]; then
+ *       CONFIG="playwright.config.ts"
+ *     fi
+ *     npx playwright test --config="$CONFIG"
+ *
+ * Note what is missing: `--project`. So whichever config it picks, EVERY
+ * project in it runs. The root config declares three:
+ *
+ *   chromium     — the regression suite. This is the one CI wants.
+ *   docs-capture — journeydoc screenshot capture (ADR-030); re-shoots every
+ *                  tutorial screenshot, and has its own dedicated job.
+ *   visual       — pixel-diff baselines, which are host-font/GPU specific and
+ *                  will not byte-match on a CI Linux runner.
+ *
+ * Letting the root config be picked would therefore run two projects that
+ * cannot pass on a runner, on top of the one that can. `playwright-test-path:
+ * tests/e2e` in the caller makes the workflow's FIRST lookup hit this file,
+ * which declares only the regression project. The root config is untouched and
+ * stays the entry point for local runs and `npm run test:e2e:docs`.
+ *
+ * The report/output paths deliberately point at the APP ROOT: the workflow
+ * uploads `<app>/playwright-report/` and `<app>/test-results/`, so with the
+ * root config's relative paths the upload step would match nothing and
+ * silently produce an empty artifact — a failing run with no report to read.
+ */
+
+import { defineConfig, devices } from '@playwright/test'
+import * as path from 'path'
+
+import { BASE_URL } from './base-url'
+
+const APP_ROOT = path.resolve(__dirname, '..', '..')
+
+export default defineConfig({
+	testDir: __dirname,
+	globalSetup: path.resolve(__dirname, 'global-setup.ts'),
+
+	// The shared workflow caps this job at `timeout-minutes: 45`, and a job the
+	// runner kills at its cap reports **cancelled** — which is NO VERDICT: not a
+	// pass, not a failure, no information. `globalTimeout` below sits inside
+	// that cap so Playwright stops first and exits non-zero WITH a tally.
+	//
+	// This suite has never run in CI, so there is no measured slowest-pass to
+	// size against. The values start at decidesk's proven settings and should be
+	// re-derived from the first green run rather than left as a guess:
+	// `retries: 0` because a retry only ever converts red to green, and a first
+	// run needs to show what is actually flaky.
+	timeout: 20_000,
+	expect: { timeout: 10_000 },
+	globalTimeout: 38 * 60_000,
+	fullyParallel: false,
+	retries: 0,
+	workers: 1,
+
+	// `github` is what makes a killed run legible: it emits a ::error::
+	// annotation per failure AS IT HAPPENS. `list` alone prints failure bodies
+	// only in an end-of-run summary, so a run cut off at the cap shows a column
+	// of ✘ marks and not one error message.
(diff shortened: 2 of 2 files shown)


rubenvdlinde:
## First run: 271 tests discovered, 169 failed — and the suite is not CI-wired

Enabling the switch did what it was meant to do: it turned an invisible gap into a measurement. **Do not merge this PR as-is** — it would take `development` red. The enable needs to land together with the fix below.

### What the run says

Every failure is the same shape:

```
Error: page.goto: Protocol error (Page.navigate): Cannot navigate to invalid URL
Call log:
  - navigating to "/apps/shillinq/accountant-portal", waiting until "load"
```

The navigation target is the **raw relative path**. `baseURL` was never applied, so every `page.goto()` and every `apiRequestContext` call is an invalid URL.

### What is NOT the cause

Checked, so the next person does not repeat it:

- **The env is present.** The Playwright step exports `BASE_URL`, `NEXTCLOUD_URL` and `NC_BASE_URL`, all `http://localhost:8080` — confirmed in the step's own env dump.
- **The resolver is correct.** `tests/e2e/base-url.ts` accepts all three names and throws when none is set. It cannot silently return empty.
- **The config sets it.** `playwright.config.ts` has `baseURL: resolveBaseURL()` in top-level `use`, and the projects only add `devices`/`storageState`, which Playwright merges rather than replaces.
- **The config path fallback exists.** The shared workflow resolves `<playwright-test-path>/playwright.config.ts` and falls back to the root `playwright.config.ts` when that is absent — and shillinq's config is at the root.

So each part looks right in isolation, and the composed result still has no `baseURL`. That is the thing to reproduce locally: run `BASE_URL=http://localhost:8087 npx playwright test --config=playwright.config.ts --list` and print `use.baseURL` from the resolved config, rather than reasoning about it further from logs.

### Why this still belongs on the record

`enable-playwright` defaults to `false` and this repo never set it, so **271 tests across 39 spec files have never executed** — while the e2e column reported `skipped`, which reads as "fine". The suite has been accumulating for months with nothing exercising it. Whatever it asserts about a real Nextcloud is unverified, and the first honest measurement is this one.

rubenvdlinde:
## Root cause of the 169 failures: `baseURL` was `undefined`

Not the env, not the resolver, not the workflow. **The config imported a name that does not exist.**

```ts
import { BASE_URL } from './base-url'   // base-url.ts has no such export
...
baseURL: BASE_URL,                       // therefore: undefined
```

`tests/e2e/base-url.ts` exports `resolveBaseURL()` and `BASE_URL_ENV_NAMES`. There is no `BASE_URL` const. **Playwright transpiles its config without type-checking**, so a non-existent named import is not an error — it is `undefined`. With `baseURL: undefined`, every relative `page.goto('/apps/shillinq/...')` becomes *"Cannot navigate to invalid URL"*.

The workflow was exporting `BASE_URL=http://localhost:8080` correctly the whole time. The resolver was correct the whole time. It was simply never called.

### Proven both directions, locally, against this repo's own config

| config | resolved baseURL | probe test |
|---|---|---|
| `import { BASE_URL }` (as it was) | `undefined` | **fails** |
| `import { resolveBaseURL }` (this fix) | `http://localhost:8087` | **passes** |

I also ruled out the plausible-but-wrong explanations before landing this: the step env did contain `BASE_URL`; the root `playwright.config.ts` resolves correctly; the workflow's `<playwright-test-path>/playwright.config.ts` → root fallback works. The reason the call log showed a bare relative path is the tell — with a real `baseURL`, Playwright logs the **full** URL, which I confirmed with a control.

### Why the type system missed it

Worth remembering rather than just fixing: `tsc` would have caught this instantly, but nothing type-checks a Playwright config on the way to running it. Any config-level import is effectively unchecked at runtime.

rubenvdlinde:
## After the `baseURL` fix: 64 passing, 83 failing — and the failures are now real

| run | outcome |
|---|---|
| before (baseURL `undefined`) | **169 failed**, 0 could navigate — every failure was `Cannot navigate to invalid URL` |
| after this fix | **64 passed**, 83 failed, 26 skipped — `Cannot navigate to invalid URL` appears **zero** times |

So the wiring is fixed: the browser now reaches the app and 64 assertions hold. What is left is the suite's real content meeting a real Nextcloud for the first time.

### What the 83 are

```
  90  element(s) not found
  88  expect(locator).toBeVisible() failed
  22  locator.waitFor: timeout 20000ms
  20  expect(received).toContain(expected)
  12  locator.click: timeout 20000ms
```

Selector- and content-level mismatches, not infrastructure. That is the expected shape for a suite that has accumulated for months with nothing running it — the specs encode an app that has since moved.

### Recommendation

**Do not merge this PR as-is.** It would take `development` red, and the 83 are a real backlog, not a quick fix. Two sane options:

1. **Land the switch with the failures quarantined** — mark the 83 `test.fixme()` with a tracking issue, so the 64 that pass become a real ratchet immediately and the rest are visible debt rather than invisible.
2. **Fix the 83 first, then flip the switch** — cleaner history, but the suite stays unrun in the meantime, which is the state that let this drift.

I would take (1): the value of this whole exercise was turning an invisible gap into a measured one, and quarantining preserves that while the backlog is worked down. That is a judgement call for whoever owns shillinq's e2e.

Either way the `baseURL` fix on this branch stands on its own and is worth keeping.

rubenvdlinde:
## Status: do not merge yet — the job would be red, and now I can say exactly why

I measured this suite against a real instance (NC 34 on `:8117`, shillinq **and** OpenRegister both built and served). Two systemic blockers sit in front of the 53 specs, and one of them is now fixed on this branch.

### Blocker 1 — the suite was asserting against a modal (FIXED here, 9119ac53)

Until `legal_country` / `legal_region` / `rgs_template` / `administration_id` are all set, the app renders **nothing but its "Set up this app" wizard**. Not a degraded page — no `#shillinq` root at all:

```js
document.querySelector('#shillinq')            // null
document.querySelector('#app-content-vue')     // null
[...document.querySelectorAll('dialog')]       // ["Set up this app 1 Welkom bij Shillinq 2 …"]
```

That is why the pre-fix run reported **77 failed / 160 passed / 25 skipped**, with `element(s) not found` (100) and `toBeVisible() failed` (96) as the dominant errors. The suite had no onboarding step, so it was describing an un-onboarded instance, not the application.

`globalSetup` now completes setup through the setup API (`POST /api/setup/config`, then the `init-administration` and `seed` actions) and marks the getting-started walkthrough seen — once setup completes, a 4-step tour opens over the dashboard and swallows clicks. Failing here is fatal by design: a suite that proceeds against an un-onboarded instance produces a red run that reads as broken application code, which is exactly the wrong diagnosis.

Verified both directions:

| control | result |
|---|---|
| every setup key deleted (`occ config:app:delete`), then run | globalSetup re-onboarded (`administration_id=ADM-001`), **4/4 passed** |
| `occ app:disable shillinq`, then run | all four calls 404 → globalSetup **throws** with the statuses in the message, 292 specs do not run |

### Blocker 2 — the seeded administration never reaches the user's context (NOT fixed, needs an owner)

`init-administration` reports success:

```json
{"success":true,"message":"Default administration created.",
 "detail":{"success":true,"seeded":0,"skipped":1,"administrationCode":"ADM-001"}}
```

…and `GET /api/setup/status` then reports `completed: true` for all five steps. But the endpoint the app actually reads returns nothing:

```
GET /index.php/apps/shillinq/api/administrations/context
{"userId":"admin","administrations":[],"activeAdministrationId":null}
```

So the admin user has **no administration in context**, and every surface gated on one stays unreachable. Concretely, `/bbv-dashboard` resolves to the catch-all redirect and renders the Financial overview instead — `getByTestId('bbv-compliance-dashboard')` is never in the DOM. That is the whole of the BBV cluster: `waterschappen-bbv-variant` (13), `provincies-bbv-variant` (9), `bbv-compliance` (3).

I tested and **refuted** the obvious explanation. The BBV menu is gated on `visibility: { field: administrationType, in: ["waterschap"] }`, so I set `legal_region=waterschap` and re-ran `waterschappen-bbv-variant` alone: still **13 failed / 1 passed**, unchanged. The organisation type is not the variable — the empty administration context is.

This one is shillinq domain behaviour, and I have deliberately not guessed at it: a fixture that forces an administration into context could easily make the suite green for the wrong reason.

### Measurement caveat

Both full-suite runs stopped short — 30 and 86 tests respectively reported as `did not run`. So **77 is a floor, not a total**, and the pass counts between the two runs are not comparable because they stopped at different points. The per-spec numbers quoted above come from targeted runs of those files alone, which did complete.

### Recommendation

Land blocker 1 (it is a strict improvement and independently verified). Then either fix blocker 2, or quarantine the administration-dependent specs with `test.fixme()` and an issue, before enabling the CI job. Enabling it as-is publishes a permanently red gate, and a gate everyone learns to ignore is worse than no gate.

rubenvdlinde:
**Correction to my analysis on this PR.** The e2e numbers I posted here — 77 failed / 160 passed, and the follow-up BBV cluster figures — are **not evidence about `development`**.

The container I measured in mounts a worktree pinned to `fix/delete-journal-posting-guard`, last commit **2026-08-12**, which is **101 commits behind `development`**. On that tree the API route `GET /bbv-dashboard` still shadowed the SPA page route of the same path and returned `412 {"message":"CSRF check failed"}` on any deep link — which is exactly what the specs do (`page.goto('/apps/shillinq/bbv-dashboard')`). `development` had already moved it to `/api/bbv-dashboard`.

Audited all of `development` for that class of defect: **590 manifest page routes vs 98 GET API routes → no collisions.**

I have also closed #618; both of its claims were wrong for the same reason, and one of them ('gated on administrationType') was additionally a measurement error on my side — I listed nav entries with `.slice(0, 25)` out of **356**, so the BBV entries were simply outside the window. `visibility: {field, in}` on a menu entry is not implemented in nc-vue at all.

The onboarding finding in globalSetup still stands on its own — but note `tests/e2e/ci-seed.sh` already performs those setup calls in CI via `playwright-seed-command`, so the gap was local-only.

Net: this repo still has no trustworthy e2e measurement. Getting one needs a container serving current `development`.

PR #570, opened 2026-08-16
fix(administrations): setup said completed:true while every user had no administration

### LINKED ISSUE #569: First-time setup reports completed:true while the user has no administration in context
`init-administration` reports success, `GET /api/setup/status` then reports `completed: true` for all five steps — and the endpoint the app actually reads returns an empty list.

## Measured

On a freshly-installed NC 34 instance with shillinq + OpenRegister built and served, driving the setup API as an authenticated admin:

```
POST /api/setup/config                        200 {"success":true}
POST /api/setup/action/init-administration    200 {"success":true,
       "message":"Default administration created.",
       "detail":{"success":true,"seeded":0,"skipped":1,"administrationCode":"ADM-001"}}
POST /api/setup/action/seed                   200 {"success":true,
       "message":"Chart of accounts and reference data seeded."}
GET  /api/setup/status                        200 {"completed":true, ...all five steps done}
```

Then:

```
GET /index.php/apps/shillinq/api/administrations/context
{"userId":"admin","administrations":[],"activeAdministrationId":null}
```

`occ config:app:get shillinq administration_id` → `ADM-001`. So the app-config says there is an administration, the setup wizard considers itself finished, and the admin user has none in context.

## Why it matters

Every surface gated on an administration stays unreachable, and the failure is silent — the route falls through to the catch-all redirect rather than erroring. `/bbv-dashboard` renders the Financial overview; `getByTestId('bbv-compliance-dashboard')` is never in the DOM.

In the Playwright suite (#564) this accounts for the whole BBV cluster: `waterschappen-bbv-variant` (13 failures), `provincies-bbv-variant` (9), `bbv-compliance` (3).

## Refuted explanation

The BBV menu is gated on `visibility: { field: "administrationType", in: ["waterschap"] }`, so the organisation type looked like the culprit. I set `legal_region=waterschap` and re-ran `waterschappen-bbv-variant` on its own: **13 failed / 1 passed**, identical to the `municipality` run. The organisation type is not the variable.

## Suspected

`seedDefaultAdministration()` returned `seeded: 0, skipped: 1` — it found an existing ADM-001 rather than creating one. If membership is granted only on the create path, then every instance where ADM-001 already exists (a re-run, a restored volume, a second admin) ends with an administration nobody is a member of. That is a guess from the response shape, not something I verified in the code.

## Ask

Decide who owns putting the seeded administration into the calling user's context, and make `setup/status` reflect it — `completed: true` while `administrations: []` is a status that reports the opposite of the user's actual state, which is what made this take a while to find.

rubenvdlinde:
Closes #569. Also restores the e2e onboarding fix that missed the #564 merge window.

## The e2e job is now enabled on `development` and needs both of these

#564 merged at 08:49 with head `549dd779` — the baseURL fix only. My onboarding commit was **not** in it (my push to that branch never landed; I reported it as pushed without verifying, which is on me). So `development` currently runs 53 Playwright specs against an app that first-time setup never completes. This PR carries that commit plus the product fix underneath it.

## Defect 1 — `findAdministration()` could never match anything

It filtered `findAll(['filters' => ['id' => $administrationId]])`. `filters` addresses the object's JSON **properties**; `id` is the ObjectEntity's own identifier, not a property. So the filter ran against a field that does not exist and returned an empty set for **every** value, valid uuids included. Measured against the live register:

```
findAll(filters.id = 'eb88a9f2-…')             -> 0 rows
findAll(filters.id = 'ADM-001')                -> 0 rows
findAll(filters.administrationCode='ADM-001')  -> 1 row
find('eb88a9f2-…')                             -> the record
```

`buildContext()` skips a membership whose administration doesn't resolve, and that path returned `null` **without logging**. Now resolved by uuid via `find()`, falling back to the `administrationCode` filter — both id spaces are in the data — and an unresolvable membership is logged rather than silently dropped.

## Defect 2 — nothing ever created an `AdministrationMembership`

Access is granted by a membership (REQ-MA-003), and no code path in the repo created one. Seeding the `Administration` record is not enough, so even with defect 1 fixed the context stays empty on a fresh install, for every user.

`init-administration` now grants the operator running setup an `eigenaar` membership on the seeded administration, written against its OpenRegister uuid. Only the **calling** user, who has already passed `AuthorizedAdminSetting`. Idempotent, and it never throws — a failure is reported in the result and logged rather than turning a successful seed into a failed setup action.

## Verified live, one request sequence

```
before  GET  /api/administrations/context
        {"administrations":[],"activeAdministrationId":null}
action  POST /api/setup/action/init-administration
        detail.membership = {"granted":true,"administrationId":"eb88a9f2-…"}
after   GET  /api/administrations/context
        administrations[0] = {administrationCode:"ADM-001", role:"eigenaar",
                              mayPostJournalEntries:true, mayCloseFiscalYear:true}
        activeAdministrationId = "eb88a9f2-…"
```

Idempotency: a second POST returns `{"granted":false,"reason":"already a member"}`, still exactly one membership, context still one administration.

⚠️ My first attempt at this measurement showed **no change at all**, because opcache (`revalidate_freq=60`) was still serving the old bytes. The tell was the response shape — `detail` had no `membership` key. Restarting the container made the fix appear. Worth knowing before concluding a PHP change did nothing.

### DIFF
--- a/lib/Service/AdministrationContextService.php
+++ b/lib/Service/AdministrationContextService.php
@@ -339,24 +339,54 @@ class AdministrationContextService {
 	}//end membershipsForUser()
 
 	/**
-	 * Fetch a single Administration record by its id.
+	 * Fetch a single Administration record by its id or its administrationCode.
 	 *
-	 * @param string $administrationId The administration id.
+	 * ⚠️ `findAll(['filters' => ['id' => …]])` MATCHES NOTHING. `filters` addresses
+	 * the object's JSON properties, and `id` is the ObjectEntity's own identifier,
+	 * not a property — so the filter is applied against a field that does not
+	 * exist and returns an empty set for EVERY value, valid uuids included.
+	 * Measured against a live register: filtering `id` on the real uuid returned
+	 * 0 rows, filtering `administrationCode` on `ADM-001` returned 1, and
+	 * `find($uuid)` returned the record.
+	 *
+	 * That was the whole of shillinq#569. buildContext() skips a membership whose
+	 * administration does not resolve, and this path returns null WITHOUT logging,
+	 * so every user's context came back `administrations: []` while first-time
+	 * setup reported `completed: true` — no error anywhere, and every
+	 * administration-scoped surface silently unreachable.
+	 *
+	 * Both id spaces are resolved because both are in the data: memberships
+	 * written against the OpenRegister uuid, and fixtures/config written against
+	 * the human `administrationCode` (the value `administration_id` app-config
+	 * holds, and the one ~20 other call sites scope on).
+	 *
+	 * @param string $administrationId The administration uuid or administrationCode.
 	 *
 	 * @return array<string,mixed>|null
 	 */
 	private function findAdministration(string $administrationId): ?array {
 		try {
 			$objectService = $this->container->get('OCA\OpenRegister\Service\ObjectService');
-			$matches = $objectService
+			$scoped = $objectService
 				->setRegister($this->register())
-				->setSchema('Administration')
-				->findAll(
-					[
-						'filters' => ['id' => $administrationId],
-						'limit' => 1,
-					]
-				);
+				->setSchema('Administration');
+
+			// Primary: the OpenRegister uuid, via the single-object lookup that
+			// actually addresses it.
+			$direct = $scoped->find($administrationId);
+			$arr = $this->asArray(row: $direct);
+			if ($arr !== []) {
+				return $arr;
+			}
+
+			// Fallback: the human administrationCode, which IS a JSON property
+			// and therefore is filterable.
+			$matches = $scoped->findAll(
+				[
+					'filters' => ['administrationCode' => $administrationId],
+					'limit' => 1,
+				]
+			);
 		} catch (\Throwable $e) {
 			$this->logger->error(
 				'AdministrationContextService: failed to load administration',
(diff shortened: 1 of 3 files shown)


rubenvdlinde:
## Every red job here is inherited from `development`, not caused by this PR

This PR changes exactly two files:

```
lib/Service/AdministrationContextService.php
lib/Service/SettingsService.php
```

No `.vue`, no workflow, no config. Attributing the reds:

| red job | actual cause | filed |
|---|---|---|
| `PHP Quality (phpcs / phpmd / phpstan / psalm)` | 508 errors on `development`, 143 of them `Doc comment for parameter "$objectService" missing`, in ten files this PR does not touch (`LandedCostAllocationService`, `PayrollWebhookController`, …). From `2c88358b` "clear the docblocks and dead code ADR-083 left behind". | #852 |
| `Lint Check` | `vue/attribute-hyphenation` and friends across `src/**/*.vue` — `App.vue`, `AdministratieSwitcher.vue`, `BBVComplianceDashboard.vue`, … This PR contains no `.vue` file. | #852 |
| `PHPUnit` (all 6 legs) | Pre-existing; the run that reported them was on `1aac8bc1`, before this branch merged `development` in. | — |
| `E2E Tests (Playwright)` | The job newly enabled by #564, red for #618 (BBV surfaces gated on an `administrationType` the seed never sets, and routes that resolve from neither URL form). | #618 |
| `Integration Tests (Newman)`, `Hydra Gates`, `Quality Report` | Already red on `development`'s own last completed run (`31926367937`). | — |

## Why the automated subset check flagged this PR anyway

The comparison was stale on **both** sides:

- the PR run it read was on `1aac8bc1`, this branch's **first** commit — before the `development` merge and before both corrections in `9b0d2bb7`;
- the `development` baseline it compared against was `31926367937` on `16e27e4f` at 04:19, which **predates** `2c88358b` (23:20 the night before). The only newer `development` run was **cancelled**, so it carries no verdict.

So the baseline was describing an older tree than the PR, and the difference got attributed here. That is worth stating plainly because it will do the same to the next PR opened against this repo until `development` has a current run.

Merging on that measurement. The two files' behaviour is verified live end-to-end (before/after context, idempotency, both id spaces), and the reds belong to `development`.

PR #851, opened 2026-08-16
fix(gate-57): remove five orphaned write capabilities

rubenvdlinde:
## gate-57 orphaned-write-capability — shillinq

Workstream **S8** of the fleet debt programme. Base at branch point `eb75115a`.

⚠️ Two things about the base, both from the coordinator's 11:25 bulletin: shillinq's `Hydra Gates` job is **`cancelled`** in the recent `development` runs, so gate parity **cannot be judged from CI** — the before/after below is the gate's own helper run locally (programme law L2). And run `31937325844` is red on **phpstan, phpmd, psalm, phpcs** from the `refactor/adr-084-type-hint-the-contract` merge — **slot R1's scope, not this PR's.**

### Gate evidence (L1 / L3 / L4)

```
python3 <hydra-gates>/scripts/lib/check_orphaned_write_capability.py <316 files> > out.txt
```
File set reproduced from `run-hydra-gates.sh`: `git ls-files -- lib/Service`, `.php`, minus `tests/`. Output captured to a file; never piped.

| | findings | files measured | helper exit |
|---|---|---|---|
| before | **6** | 316 | 0 |
| after | **1** | 316 | 0 |

Helper identity line: `app_id=shillinq source=appinfo/info.xml foundation=no`.

⚠️ **shillinq does not vendor `conduction/hydra-gates`** (no `conduction/` directory in `vendor/` at all), so the helper and the phpcs/phpmd rulesets were mounted in from docudesk's vendored copy. Without that, `vendor/bin/phpcs` aborts on a missing ruleset — a tool that never loaded its rules is not a clean result.

### Per-finding judgement — 6 findings: 5 deletes, 0 wires, 1 escalated

This is the app gate-57 was written from, and all five removals are the exact shape its docstring describes: *"fully implemented, unit-tested by calling the class directly, and spec'd done — with ZERO production callers."*

| # | finding | judgement | reasoning |
|---|---|---|---|
| 1 | `PayrollService.php:399 persistLoonStrook` | **DELETE** | `PayrollController::loonstrook` drives `berekenLoonStrook()` and returns the computed payload; nothing ever handed it back for a save, so no `LoonStrook` object was ever written through this service. |
| 2 | `PayrollService.php:416 persistLoonjournaalpost` | **DELETE** | Same, via `PayrollController::journaalpost` → `bouwLoonjournaalpost()`. |
| 3 | `PayrollJaaropgaveService.php:158 persistJaaropgave` | **DELETE** | Same, for the yearly aggregate. The service builds a Jaaropgave and returns it. |
| 4 | `SubsidieVerantwoordingService.php:194 persistChange` | **DELETE** | A generic *"save this payload into that register and schema, unless a record with this business key exists"* pass-through, with no caller. The two builders above it are pure and return their payloads; nothing took one back. The signature is also the widest possible write surface for the narrowest possible need — a real caller should reach OpenRegister's ObjectService for its own schema (ADR-022), not through a pass-through that accepts an arbitrary target. |
| 5 | `StatementManifestService.php:92 importForced` | **DELETE** | `run(force: true)` — a one-statement variant of `import()` whose whole purpose is to **overwrite operator-edited manifests**. Its docblock claimed "used by forced register upgrades"; no such caller exists, and the only consumer, `Repair\InitializeSettings`, calls `import()`. A destructive capability waiting for its *first* caller, not one that lost its last. |
| 6 | `DunningRunService.php:926 transferToIncasso` | **ESCALATED — left in place** | See below. |

**The two guards that go with the payroll deletions — read this before approving.** `persistLoonjournaalpost()` refused an unbalanced journal (REQ-PAY-012) and `persistJaaropgave()` refused inconsistent cumulatieven (REQ-PAY-013). **Neither guard ever executed, because neither method ever ran.** Removing them takes away nothing that was protecting anything — but that is exactly the point worth recording: an invariant that must hold belongs on the `Loonjournaalpost` / `Jaaropgave` **schema in OpenRegister**, where it binds every writer, rather than on a service method with no callers where it protects only the unit test that calls it. That is on the fleet board as a follow-up. The balance invariant itself is still asserted, on the payload the live controller path actually returns, by `testBouwLoonjournaalpostBalances` and `testBouwJaaropgaveFlagsCumulatievenMismatch`.

### The one finding I did not fix, and why (L8)

**`DunningRunService::transferToIncasso` is left exactly as it is.** Neither deleted nor wired.

- A **live spec requires the capability**: `openspec/specs/bookkeeping-credit-control-dunning/spec.md` REQ-CCD-008 — *"Optionele overdraft-actie stage 5 SHALL POST dossier-bundel to incassobureau-API"*. `DunningController::dossier` composes the bundle (`IncassoDossierComposer::compose`) and returns it; nothing sends it, and `transferToIncasso` is the method that would seal the run when the incasso adapter reports DELIVERED.
- Wiring it means the dossier endpoint would actually hand a debt to a collection agency. That is a product decision with a real-world effect on a real debtor, not debt cleanup, and it is not mine to make.

So gate-57 stays at **1 finding** on shillinq. Under programme law L8 a red gate is worth more than a green one that tests nothing, and I would rather hand this back than close it with an exclusion comment. Recorded on the fleet board for a decision.

**No gate-exclusion comment was used anywhere in this PR.**

### Quality (L10)

Run under PHP 8.3 in a `php:8.3-cli` container against this worktree (host PHP is 8.2; shillinq's `vendor/` requires >= 8.3 and phpcs exits **255** on the host, which is an abort, not a pass):

- `lint` — clean on every changed file
- `phpcs` on the 4 changed `lib/` files — **0 errors**. That includes **two pre-existing phpcs errors fixed here** (`Doc comment for parameter "$objectService" missing`, in `PayrollService` and `PayrollJaaropgaveService`), per the fix-what-you-touch rule; both files are now phpcs-clean.
- `phpmd` on the same 4 files — clean, **exit 0**. This PR only deletes, so no complexity ceiling moves.
- `phpunit` — the branch point is already red on the adr-084 `ObjectServiceInterface` rollout (R1). Please compare this PR's own CI to run `31937325844` job by job rather than to a green expectation.

Unused-import cleanup that came with the deletions: `RuntimeException` in `PayrollJaaropgaveService` and in its test.


rubenvdlinde:
**S33: rebased onto `development`; `dirty` cleared, `mergeable=MERGEABLE`.**

**The predicted overlap did not exist.** #853 touches only `lib/Controller`; this PR touches only `lib/Service` + `tests/Unit/Service`. They never met.

**Both conflicts were the same docblock line**, written independently on each side for the same phpcs *"Doc comment for parameter `$objectService` missing"* left by the ADR-084 rollout:

```
lib/Service/PayrollService.php:61
lib/Service/PayrollJaaropgaveService.php:60

ours    @param ObjectServiceInterface $objectService OpenRegister object service.
theirs  @param ObjectServiceInterface $objectService OpenRegister's object service, injected per ADR-083.
```

**Resolution, per conflict: `development`'s wording, both times** — it is the merged convention, it names the ADR the parameter exists for, and neither side carries behaviour, so there is nothing to weigh but wording. Applied with git's own `-X theirs`; **no hand-edit stands in either file**.

**#853's security fix survives intact — verified, not assumed:**
- All four controllers it hardened (`ARInvoiceEInvoiceController`, `PayrollController`, `PeriodCloseController`, `SoftCloseController`) are **byte-identical to `origin/development`** in the merged tree.
- None of the five methods deleted here is a method #853 guarded.

**All five deletions survive and `development` reintroduced no caller**: `persistJaaropgave`, `persistLoonStrook`, `persistLoonjournaalpost`, `importForced`, `persistChange` — zero definitions, zero call sites across `lib/`, `src/`, `tests/`.

**gate-57 re-measured after the merge** — `.github@main` `check_orphaned_write_capability.py` (1094 lines), **316** `lib/Service` files each side, `foundation=no`, rc=0 both, output to a file:

```
BASE  origin/development 29b9e98c : 6 findings
HEAD  merged branch      e178fd0c : 1 finding  — DunningRunService::transferToIncasso
```

The survivor is the one escalated on purpose. **6 → 1 reconfirmed after the merge**, not carried forward from before it.

The branch has now fully absorbed `development`: **two-dot and three-dot diffs are identical** (8 files, +66/−310), so the delta gates can no longer attribute the base's movement to this PR.

PR #854, opened 2026-08-16
fix(deps): gate-66 ADR-083 — the twelve OpenRegister lookups #556 left behind, two of them fatal

rubenvdlinde:
### What this is, and what the "236" in the fleet board was

The gate matrix lists shillinq at **236 ADR-083 violations**. That number was
real, and it was measured against a tree that `development` has since left
behind. Same checker, four trees:

| tree | FAIL lines |
|---|---:|
| `16e27e4f` — the sha behind the §1 baseline run 31926367937 | **236** |
| `191156f8` — corrected-baseline run 31937325844 | 12 |
| `49968ce0` — this PR's base | **12** |

**224 of the 236 were already fixed by #556** (`refactor/adr-084-type-hint-the-contract`,
merged 08:46:48), which converted 148 files under `lib/` to constructor
injection. This PR is the residue #556 left behind — and part of that residue
is a live defect, not a style finding.

### The systemic pattern — one shape, two sub-shapes

**A — `$this->container` survived a constructor that no longer has a container.**
#556 removed `ContainerInterface $container` from three services and converted
their call sites, but missed one site in each. Those classes now declare **no
container property and no `use Psr\Container\ContainerInterface`**, so
`$this->container->get(...)` is `Error: Call to a member function get() on null`
the moment the method runs:

| site | reachable |
|---|---|
| `lib/Service/AansluitingService.php:681` | gate-reported |
| `lib/Service/DoorsnijdingsVerbodValidator.php:230` | gate-reported |
| `lib/Service/IcpFilingService.php:303` | **not gate-reported — see below** |

`IcpFilingService::saveReturn()` is the one that matters. Its dangling lookup
sits inside `catch (\Throwable) { return false; }`, so **the method has been
returning `false` on every call — an ICP correction filing is accepted and
silently never persisted.** gate-66 correctly does not report it: a degrading
`catch` is precisely ADR-083 rule 1's optional-capability exception. **The catch
that makes the lookup legal is the same catch that swallows the fatal.** It was
found by sweeping `lib/` for `$this->container` in files that declare no
container, which is a check gate-66 does not and should not perform:

```sh
for f in $(grep -rl 'this->container' lib/); do
  grep -qE '(ContainerInterface|IAppContainer|ContainerBuilder)\s+\$container|\$container;' "$f" \
    || echo "DANGLING: $f"
done
```

The sweep returns four files on this repo; the fourth,
`lib/Reporting/Generator/ReportDataTrait.php`, is a legitimate trait contract —
all seven generators that `use` it do declare `$container` — and is untouched.

**B — eight classes #556 never visited** (10 of the 12 findings). They still take
`private readonly ContainerInterface $container` and pull ObjectService
mid-method. In every one of the eight, **that lookup is the container's only
use**, so the parameter comes out and `ObjectServiceInterface $objectService`
goes in, matching the 148 classes already converted and the
`registerServiceAlias(ObjectServiceInterface::class, 'OCA\OpenRegister\Service\ObjectService')`
already in `lib/AppInfo/Application.php:186`.

All eight are autowired (`registerEventListener(listener: X::class)` /
container-resolved), so no registration closure changes.

### Not applicable here

- **Rule 2** (OpenRegister type in a class header): no findings.
- **Rule 3** (default route): the checker prints
  `NOTE: no default route found in appinfo/routes.php and no #[FrontpageRoute] attribute, so rule 3 inspected nothing.`
- **The `class_exists()`-in-`register()` hazard** flagged on hermiq does not
  exist here: `Application.php:186` aliases **two strings**, so nothing
  autoloads at registration and the plumbing cannot silently skip.

### Evidence

**gate-66**, run exactly as CI runs it — `run-hydra-gates.sh:10099` invokes
`python3 lib/check_openregister_dependency_shape.py .` from the repo root and
counts `^FAIL ` lines. shillinq's `code-quality.yml` deliberately sets **no
`hydra-gates-ref`**, so CI uses `.github@main`; the checker was fetched from
there by `download_url` + `curl` and verified as blob **`8db55c4c…`, 18 979
bytes, 453 lines** against both `contents/…?ref=main` and an independent
`git clone --depth 1 --branch main`.

| | findings | files | exit |
|---|---:|---:|---:|
| **before** (`49968ce0`) | **12** | 680 | 1 |
| **after** | **0** | 680 | 0 |

gate-66 is **full-scope by construction** — its own runner block says *"NOT
diff-scoped. The shape of a dependency is a property of the tree, not of a
diff"* — so before and after are the same 680 files and are directly
comparable. The checker demonstrably can fail: the pre-`.github#458` generation
of it reports 206 on this same tree.

**PHP quality**, PHP 8.3 container, on the 16 files this PR touches, base
(`49968ce0`) vs head:

| tool | base | head |
|---|---:|---:|
| `composer lint` (php -l, whole repo) | pass | **pass** |
| `phpcs --standard=phpcs.xml` | 96 errors / 1 warning | **52 errors / 1 warning** |
| `phpmd` (both configs) | 0 | **0** |

All **eleven `lib/` files are now phpcs-clean (0 errors)** — that includes
clearing the docblock debt #556 left behind: three constructors missing
`@param $objectService`, plus the orphaned continuation lines of the deleted
`@param ContainerInterface` (`*    lazily.`), which tripped
`Generic.Commenting.DocComment.LongNotCapital`.

The residual 52 are **pre-existing and unchanged in count**, in two test files:
`ReconciliationResolutionServiceTest` 42 (base 42) and
`SignoffDecisionConcludedListenerTest` 8 (base 20 — improved), all
`CustomSniffs.Functions.NamedParameters` on PHPUnit assertions and missing
docblocks on a `class_exists` polyfill stub. They are untouched deliberately:
`PHP Quality` is red on this base for the ADR-084 rollout, that leg is another
slot's, and rewriting 42 assertion call sites here would collide with it.

**Tests.** The five unit classes that construct the changed classes:

```
StatementVerifyGuardVarianceTest            2/2
ReconciliationResolutionServiceTest         8/8
ACMReportSignTransitionListenerTest         6/6
AnnualReportSignoffRequestListenerTest    ) 11/11
SignoffDecisionConcludedListenerTest      )
```

Two of them were handing the service a **plain array where the real contract
returns `?ObjectEntityInterface`**. The untyped `ContainerInterface::get()` was
what let that shape through — nothing enforced the return type of a fake pulled
out of a container. Injecting the typed contract makes those doubles honest, and
they now return `ObjectEntity` stubs the service normalises through its own
`toArray()`, exactly as it does in production.

### Parity

`Hydra Gates` is **`cancelled`** in this base's run (31937325844), and
`GET /actions/runs/<id>/jobs?per_page=100` currently returns `total_count: 0`
for every recent shillinq run — unreadable, which is not green. Gate parity is
therefore asserted from the gate's own script locally, on both the base commit
and this head, at the same 680-file scope.

The four `PHP Quality` legs (phpcs, phpmd, psalm, phpstan) are red on this base
from the ADR-084 rollout and are not this PR's; this PR adds no new failing job.

### A third residue, in the one file gate-66 skips on purpose

`lib/AppInfo/Application.php:387` still constructed `DoorsnijdingsVerbodValidator`
with `container: $c` — a **named argument for a parameter #556 deleted**. That is
`Error: Unknown named parameter $container` on every resolution, and it never
passed the `ObjectServiceInterface` the constructor now requires: the
doorsnijdingsverbod check (REQ-IBA-004) could not be constructed at all.

gate-66 cannot report this, correctly — its checker skips the composition root
by name (*"`lib/AppInfo/Application.php` is the COMPOSITION ROOT … Rule 1 has
nothing to say about it. 16 of 435 findings sat here"*). That exclusion is right
for rule 1, and it is why a constructor-signature change needs a separate check
of the wiring layer:

```sh
grep -n "container: \$c\|container: \$container" lib/AppInfo/*.php
```

Three hits here; the other two (`FourEyesPaymentRunGuard`,
`PaymentRunDuplicateGuard`) are consistent — both classes still declare
`ContainerInterface $container` — and are untouched. **The named argument is what
made it silent**: positionally PHP would have thrown a type error on argument #1;
by name it fails only on the name, at resolution time, on a path no unit test
constructs.

### One measured red → less red, and one that is honestly still red

`IcpFilingServiceTest::testCreateCorrectionReattachesEvidence` fails on the base
at line 252, `assertTrue($result['saved'])` — *"Failed asserting that false is
true"* — which is the `saveReturn()` defect above, pinned by a test that has been
red. On this head it gets past 252 (8 → 9 assertions) and fails later, at line
254, on a **different** cause: `buildService()` wires its recording stub into
`$this->container->method('get')`, which nothing consumes any more, and hands each
service a bare `createMock(ObjectServiceInterface::class)` instead. The fixture is
disconnected from the code under test.

That is a fourth shape of the same refactor's fallout — **a test double orphaned
from its subject** — and it is deliberately *not* fixed here: it belongs to the
ADR-084 test fallout being worked in another PR, and 224 test files in this repo
still stub `container->method('get')`, so it wants one owner and one pass, not a
drive-by. This PR neither fixes nor worsens those three assertions; the two tests
that fail identically on base and head are untouched.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -385,8 +385,8 @@ class Application extends App implements IBootstrap {
 			DoorsnijdingsVerbodValidator::class,
 			static function (ContainerInterface $c): DoorsnijdingsVerbodValidator {
 				return new DoorsnijdingsVerbodValidator(
-					container: $c,
 					appConfig: $c->get(IAppConfig::class),
+					objectService: $c->get(ObjectServiceInterface::class),
 					auditLogger: $c->get(InnovatieboxAuditEventLogger::class),
 				);
 			}
--- a/lib/Guard/StatementVerifyGuard.php
+++ b/lib/Guard/StatementVerifyGuard.php
@@ -47,9 +47,9 @@ declare(strict_types=1);
 
 namespace OCA\Shillinq\Guard;
 
+use OCA\OpenRegister\Contract\ObjectServiceInterface;
 use OCA\Shillinq\AppInfo\Application;
 use OCP\IAppConfig;
-use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -70,22 +70,20 @@ use Psr\Log\LoggerInterface;
  */
 class StatementVerifyGuard {
 	/**
-	 * Construct the guard with lazy DI of OR's ObjectService.
+	 * Construct the guard with OR's ObjectService injected (ADR-083 rule 1).
 	 *
-	 * @param ContainerInterface $container DI container — OR's ObjectService
-	 *                                      is fetched lazily so this class
-	 *                                      stays usable without a hard
-	 *                                      compile-time dependency on
-	 *                                      OpenRegister.
 	 * @param IAppConfig $appConfig App config for dynamic register
 	 *                              slug resolution.
 	 * @param LoggerInterface $logger Nextcloud logger for fail-closed
 	 *                                diagnostics.
+	 * @param ObjectServiceInterface $objectService OpenRegister's published
+	 *                                             object surface (ADR-084),
+	 *                                             aliased in Application.php.
 	 */
 	public function __construct(
-		private readonly ContainerInterface $container,
 		private readonly IAppConfig $appConfig,
 		private readonly LoggerInterface $logger,
+		private readonly ObjectServiceInterface $objectService,
 	) {
 	}//end __construct()
 
@@ -300,8 +298,6 @@ class StatementVerifyGuard {
 			return 0;
 		}
 
-		$objectService = $this->container->get('OCA\\OpenRegister\\Service\\ObjectService');
-
 		// Page through GLLine entries filtered by accountNumber + entryDate
 		// window. We use accountNumber == bankAccountId by convention; T4
 		// expects bank accounts to share their IBAN/identifier with the GL
@@ -312,7 +308,7 @@ class StatementVerifyGuard {
 		$page = 1;
 		$batchSize = 0;
 		do {
-			$batch = $objectService
+			$batch = $this->objectService
 				->setRegister($this->getRegisterSlug())
 				->setSchema('GLLine')
 				->findAll(
@@ -366,8 +362,7 @@ class StatementVerifyGuard {
 	 */
 	private function persistVarianceFields(string $id, int $expectedCents, int $varianceCents): void {
 		try {
-			$objectService = $this->container->get('OCA\\OpenRegister\\Service\\ObjectService');
-			$objectService
+			$this->objectService
 				->setRegister($this->getRegisterSlug())
 				->setSchema('BankReconciliation')
 				->updateObject(
@@ -398,14 +393,12 @@ class StatementVerifyGuard {
 	 * @return array<int, array<string, mixed>> The match object arrays.
 	 */
 	private function findMatches(string $reconId): array {
-		$objectService = $this->container->get('OCA\\OpenRegister\\Service\\ObjectService');
-
 		$matches = [];
 		$pageSize = 500;
 		$page = 1;
 		$batchSize = 0;
 		do {
-			$batch = $objectService
+			$batch = $this->objectService
 				->setRegister($this->getRegisterSlug())
 				->setSchema('ReconciliationMatch')
 				->findAll(
--- a/lib/Listener/ACMReportSignTransitionListener.php
+++ b/lib/Listener/ACMReportSignTransitionListener.php
@@ -52,13 +52,13 @@ declare(strict_types=1);
 
 namespace OCA\Shillinq\Listener;
 
+use OCA\OpenRegister\Contract\ObjectServiceInterface;
 use OCA\OpenRegister\Event\ObjectTransitionedEvent;
 use OCA\Shillinq\Service\ListenerSchemaResolver;
 use OCA\Shillinq\Service\SettingsService;
 use OCA\Shillinq\Service\Signing\SigningDelegationService;
 use OCP\EventDispatcher\Event;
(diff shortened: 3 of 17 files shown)


rubenvdlinde:
### ⚠️ This PR cannot be verified in CI, and neither can any other shillinq PR right now

`Code Quality` did not run on this branch, and it is not a dispatch quirk:

```
$ gh workflow run code-quality.yml --ref S15/gate-66-adr-083-residue
HTTP 422: failed to parse workflow:
  (Line: 190, Col: 7): 'enable-playwright' is already defined,
  (Line: 191, Col: 7): 'playwright-test-path' is already defined
```

`development`'s own `.github/workflows/code-quality.yml` declares
`enable-playwright` and `playwright-test-path` **twice** in the same `with:` block
(lines 133/134 and 190/191). The file does not parse, so GitHub opens a run, finds
no jobs and marks it `completed`. That is why recent shillinq runs return
`total_count: 0` jobs, why `Hydra Gates` reads as "cancelled" when it in fact never
started, and why this PR has 6 check-runs instead of ~40.

Two branches enabled Playwright independently and the merge kept both blocks —
`820088c2` had one, the merge `79eaf875` had two, and it reached `development`
via #440 after #564 had already added the first. A duplicate key is invisible in a
merge diff: both sides added a line that was not there before.

**Deliberately not fixed here.** The two blocks are not identical — one carries the
`tests/e2e/playwright.config.ts` selection rationale, the other the `additional-apps`
OpenRegister pin and the `pre-test-command` — so deleting the wrong one silently
changes which Playwright config CI uses. And the moment it parses,
`enable-playwright: true` takes effect for the first time: 53 specs that have never
run. That belongs with whoever owns #440/#564, not in a dependency-shape PR.

The evidence in the description above is therefore all **local**, measured on both
the base commit and this head with the same tools at the same scope.

PR #857, opened 2026-08-16
fix(innovatiebox): the validator factory still passed the $container ADR-084 removed

rubenvdlinde:
## The defect — a second live 500, found by sweeping every composition root in the fleet

`Application.php:387` registers `DoorsnijdingsVerbodValidator` with a hand-written factory that passes `container: $c`. ADR-084 (#556) **replaced that constructor parameter** with the published `ObjectServiceInterface` contract:

```php
public function __construct(
    private readonly IAppConfig $appConfig,
    private readonly ObjectServiceInterface $objectService,
    private readonly ?InnovatieboxAuditEventLogger $auditLogger = null,
) {
```

So resolving the service throws `Error: Unknown named parameter $container` — and `$objectService`, which replaced it, is never passed at all. `InnovatieboxController::__construct()` requires the validator (`lib/Controller/InnovatieboxController.php:78`), so this is not one endpoint: **every Innovatiebox route is a 500**.

**Nothing we run notices this.** `php -l` passes. phpcs, phpmd, psalm and phpstan pass. `tests/Unit/Service/DoorsnijdingsVerbodValidatorTest` passes — it constructs the validator itself, with the *correct* two arguments, and never goes through the factory. The first thing that notices is a request.

⚠️ And this repo had **no CI at all** between ~09:32 and #856, so there was nothing to notice it with either. A repo that cannot be measured cannot report what broke in it.

## How it was found

Not by grep. Every `new <Class>(...)` in all 18 repos' `lib/AppInfo/Application.php` was compared against the target class's real `__construct()`, parsed from that repo's own tree — **260 call sites, 18 repos, each at its `origin/development` tip**. Two repos had a mismatch: softwarecatalog (`/me`, PR #531) and this one.

## The test

`tests/Unit/AppInfo/CompositionRootArgumentsTest.php` closes the hole for **every** hand-written factory here, not only this one. It tokenises `Application.php` (so comments and strings cannot be mistaken for code), resolves each class name the way PHP does — imports first, then the file's own namespace — and reflects the target constructor.

**Shown to fail against the unfixed composition root** (`git stash` of the one hunk, same command, same container):

```
1) CompositionRootArgumentsTest::testEveryManualFactorySuppliesAllRequiredConstructorArguments
  - Application.php:387 new DoorsnijdingsVerbodValidator — named argument $container
    does not exist on DoorsnijdingsVerbodValidator::__construct() (it takes:
    $appConfig, $objectService, $auditLogger). PHP fatals with "Unknown named
    parameter $container".
  - Application.php:387 new DoorsnijdingsVerbodValidator — required constructor
    parameter $objectService is not supplied (3 of 3 arguments given).

2) CompositionRootArgumentsTest::testInnovatieboxValidatorFactoryMatchesItsConstructor
The DoorsnijdingsVerbodValidator factory passes named argument(s) $container that
its constructor does not have. Resolving the service throws "Unknown named
parameter", and InnovatieboxController requires it — so every Innovatiebox route
returns 500.

Tests: 4, Assertions: 14, Failures: 2.
```

With the fix: `OK (4 tests, 15 assertions)`.

It refuses to report a vacuous pass — a check that did not run looks exactly like one that passed:

- a **floor of 20 parsed call sites**, and `DoorsnijdingsVerbodValidator` must be among them;
- a **floor of 5 reflected constructors**;
- `testUnresolvableFactoryTargetsAreDeclared` keeps the set of targets that cannot be reflected in a unit-test run **explicit**. It is currently **empty** — all 28 call sites are actually checked — and a new entry appearing there fails the test rather than silently shrinking the check.

(That last guard earned its place during development: the first version of the parser did not resolve unqualified names against the file's own namespace, and reported `OpdrachtUitvoeringGateRegistration` and `SigningDelegationRegistration` as unreflectable when both are simply neighbours of `Application`. Unresolvable reads as unchecked, and unchecked reads as clean.)

## Before / after

| measurement | before | after |
|---|---|---|
| composition-root argument mismatches in `lib/AppInfo/Application.php` | **2** | **0** |
| `new <Class>(...)` call sites scanned, both sides | 28 | 28 |
| factory targets that could not be reflected | 0 | 0 |
| `CompositionRootArgumentsTest` | 2 failures | 4 tests, 15 assertions, **OK** |

Instrument output redirected to a file and line-counted — never piped, so no exit code is read off the last stage of a pipe. Run identically on both sides; the only difference is the one hunk.

## Quality legs, on the changed files

- `phpcs --standard=phpcs.xml lib/AppInfo/Application.php` — **rc 0**, 0 errors, 3 pre-existing `@spec` warnings on `Application`/`register()`/`boot()` (untouched by this PR).
- `phpstan analyse lib/AppInfo/Application.php tests/Unit/AppInfo/CompositionRootArgumentsTest.php` — **`[OK] No errors`, 2 files.** Positive control in the same container: `phpstan analyse lib/Lifecycle` reports **15 errors**, so the instrument can say no here.
- `php -l` clean on both files.

Container: `hermiq-llm-runner:gates` (PHP 8.3.33) — the host is PHP 8.2, and `php:8.3-cli` ships neither `python3` nor `git`.

## Parity

Base `6bd201f0`, which is the first commit after **#856 repaired this repo's CI** (a duplicate key in one `with:` mapping meant GitHub could not parse `code-quality.yml`; those runs reported `failure` with **zero jobs**). Any shillinq run created between ~09:32 and #856 means nothing — **count the jobs before believing this PR's run, and before believing any base you compare it to.**

The repo's inherited red on this base is the ADR-084 fallout measured elsewhere: 4 PHP Quality legs and **506 failing tests** (`Tests: 4133, Errors: 162, Failures: 344`, measured by S4 on a detached worktree at the base sha). This PR touches one factory and adds one test file; it cannot reduce that set, and it does not add to it.

## Not done here, deliberately

- The **506 broken tests** and the phpcs/phpstan `$objectService` docblock family are service-layer and call-site work, a different change from the composition root. Keeping this PR to the runtime fatal makes it safe to merge immediately, which is the point — it is live right now.
- `DoorsnijdingsVerbodValidator`'s constructor docblock still carries the orphaned fragment left when `$container`'s `@param` was removed, and has no `@param` for `$objectService`. Left alone: it is one instance of a repo-wide phpcs family that belongs to whoever takes that sweep, and touching it here would blur what this PR is.

---

## Added after opening: the base measured, S7's way

`origin/development` at `6bd201f0`, extracted read-only into a detached tree, unit suite run locally against this repo's own `vendor/` in `hermiq-llm-runner:gates`:

```
Tests: 4133, Assertions: 42219, Errors: 162, Failures: 346,
Warnings: 32, Deprecations: 1, Skipped: 1
```

That independently reproduces S4's `4133 / 162 / 344` at its branch base (the two extra failures are the merges since), so **508 broken tests are inherited on this base and none of them is this PR's.** Dominant error signatures, all ADR-084 call-site fallout:

| count | signature |
|---:|---|
| 20 | `Named parameter $objectService overwrites previous argument` |
| 16 | `Unknown named parameter $container` |
| 16 | `Argument ... must be of type Psr\Container\ContainerInterface, null given` |
| 9 | `Return value must be of type array, null returned` |

⚠️ This is the **unit** layer (`phpunit-unit.xml`, outside Nextcloud), not the six PHPUnit matrix cells CI runs. It isolates the ADR-084 damage; it is not a substitute for CI's verdict.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -385,8 +385,15 @@ class Application extends App implements IBootstrap {
 			DoorsnijdingsVerbodValidator::class,
 			static function (ContainerInterface $c): DoorsnijdingsVerbodValidator {
 				return new DoorsnijdingsVerbodValidator(
-					container: $c,
 					appConfig: $c->get(IAppConfig::class),
+					// ADR-084 replaced the `$container` parameter with the
+					// published contract. This factory bypasses autowiring, so
+					// it does not follow a constructor automatically: passing
+					// `container:` here was `Error: Unknown named parameter
+					// $container` on every request that resolved the service —
+					// i.e. every InnovatieboxController route. See
+					// tests/Unit/AppInfo/CompositionRootArgumentsTest.php.
+					objectService: $c->get(ObjectServiceInterface::class),
 					auditLogger: $c->get(InnovatieboxAuditEventLogger::class),
 				);
 			}
--- a/tests/Unit/AppInfo/CompositionRootArgumentsTest.php
+++ b/tests/Unit/AppInfo/CompositionRootArgumentsTest.php
@@ -1,0 +1,587 @@
+<?php
+
+/**
+ * Composition-root argument tests.
+ *
+ * `lib/AppInfo/Application.php` registers a number of services with a
+ * hand-written factory closure — `return new Foo(bar: $c->get(...))` —
+ * instead of letting Nextcloud autowire them. That is a deliberate choice
+ * (several of these need an argument the container cannot infer, or must stay
+ * a string reference so a disabled OpenRegister does not fatal bootstrap), but
+ * it has one dangerous property:
+ *
+ *   **A manual `new Foo(...)` that no longer matches `Foo::__construct()` is
+ *   invisible to every static check we run.** `php -l` passes, phpcs, phpmd,
+ *   psalm and phpstan pass, and the class's own unit test passes too, because
+ *   the unit test constructs it directly and never goes through the factory.
+ *   The first thing that notices is a **request**.
+ *
+ * That is exactly what happened when ADR-084 replaced
+ * `DoorsnijdingsVerbodValidator`'s `$container` parameter with the published
+ * `ObjectServiceInterface` contract. The factory in `Application.php` kept
+ * passing `container: $c` — a parameter that no longer exists — so resolving
+ * the service threw `Error: Unknown named parameter $container`, and
+ * `InnovatieboxController` requires that service, so **every Innovatiebox
+ * route returned 500**. `tests/Unit/Service/DoorsnijdingsVerbodValidatorTest`
+ * was green throughout: it builds the validator itself, correctly.
+ *
+ * These tests close that hole. They read `Application.php`'s own source,
+ * extract every `new <Class>(...)` in it, and compare the call against the
+ * target class's real constructor via reflection.
+ *
+ * @category  Test
+ * @package   OCA\Shillinq\Tests\Unit\AppInfo
+ * @author    Conduction b.v. <info@conduction.nl>
+ * @copyright 2026 Conduction B.V.
+ * @license   EUPL-1.2 https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ * @link      https://github.com/ConductionNL/shillinq
+ */
+
+declare(strict_types=1);
+
+namespace OCA\Shillinq\Tests\Unit\AppInfo;
+
+use OCA\Shillinq\Service\DoorsnijdingsVerbodValidator;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+
+/**
+ * Every hand-written factory in the composition root must still match the
+ * constructor it calls.
+ */
+class CompositionRootArgumentsTest extends TestCase {
+
+	/**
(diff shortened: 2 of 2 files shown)


rubenvdlinde:
**Parity note on the two red non-`Code Quality` checks, measured rather than argued:**

- **`l10n`** — `completed/failure` on `development` at **`6bd201f0`**, which is this PR's exact base (run 31944518xxx-era, and on the three commits before it too). Inherited.
- **`lint-check`** — every file it reports is under `src/`: `App.vue`, `components/AdministratieSwitcher.vue`, `components/BudgetBBVMapping/*`, `components/Dashboard/BBVComplianceDashboard.vue`, `components/cashflow/*`, `components/dashboard/financial/*`, `components/goods-receipt-note/GoodsReceiptNoteForm.vue`, … **This PR touches two files, both PHP** (`lib/AppInfo/Application.php`, `tests/Unit/AppInfo/CompositionRootArgumentsTest.php`) and **zero files under `src/`**. A PR cannot cause an eslint error in a file it does not touch. This is the repo-wide eslint backlog already recorded as failing on every PR that touches `src/` — and it fails here without one.

`Code Quality` is still queued (runner congestion). **Count its jobs before believing it** — this repo produced `jobs=0` `failure` runs until #856 repaired the workflow file, and a zero-job run is unreadable, not red.

PR #858, opened 2026-08-16
test(e2e): disable Nextcloud's firstrunwizard — 84 of 90 click timeouts were its modal

### LINKED ISSUE #618: BBV variant surfaces unreachable: gated on an administrationType the seed never sets, and the routes resolve from neither URL form
Splitting this out of #569, which is fixed separately in #570. With the administration context working, the BBV variant surfaces are still unreachable — for two further reasons, both measured.

## 1. The variant surfaces are gated on `administrationType`, which the default seed does not set

`src/manifest.d/bookkeeping-waterschappen-bbv-variant-04-manifest-routes.json` gates its menu on:

```json
"visibility": { "field": "administrationType", "in": ["waterschap"] }
```

`lib/Settings/seeds/administraties/default.json` has no `administrationType` property at all — it carries `legalForm: "bv"`. So on every default install the gate evaluates against a missing field and every BBV entry stays hidden.

Measured: `PUT` the seeded ADM-001 with `administrationType: "waterschap"` and the entries appear immediately —

```
cn-nav-entry-BBVComplianceDashboard   /apps/shillinq/bbv-dashboard
cn-nav-entry-BbvComplianceDashboard   /apps/shillinq/bbv-provincie/compliance-dashboard
cn-nav-entry-BudgetBBVMappings        /apps/shillinq/budget-mappings
cn-nav-entry-BudgetToProgrammeLinker  /apps/shillinq/bbv-provincie/budget-to-programme
cn-nav-entry-BbvMapping               /apps/shillinq/overheid/bbv-mapping
```

Note this is **not** the `legal_region` app-config, which is the intuitive suspect. I tested that first: setting `legal_region=waterschap` left `waterschappen-bbv-variant` at 13 failed / 1 passed, unchanged. The gate reads a property on the Administration **object**.

Two ids also collide — `BBVComplianceDashboard` and `BbvComplianceDashboard`, from the waterschap and provincie fragments, both labelled "BBV Compliance Dashboard" and both visible at once.

## 2. Even then, the routes do not resolve from a URL

The nav emits **path**-form hrefs, but a full load of one 412s (no server route serves the SPA shell for deep sub-paths):

```
GET /index.php/apps/shillinq/bbv-dashboard   -> 412 Precondition failed
```

while the hash form the e2e specs deep-link falls through to the catch-all and renders the Financial overview:

```
#/bbv-dashboard  -> Dashboard surface, [data-testid="bbv-compliance-dashboard"] never present
```

So the surface is reachable only by clicking the nav entry in an already-loaded SPA — not by bookmark, refresh, middle-click, or deep link. openregister hit exactly this and settled on hash mode precisely because the backend registers one frontend route; the nav hrefs here look like they were authored for history mode.

## Impact

This is the remainder of the red Playwright job now enabled on `development`. After #570 lands, the measured residue on these five spec files is **40 failed / 7 passed of 47**, dominated by the BBV cluster.

## Ask

Decide (a) whether the default administration should carry an `administrationType`, or whether the BBV specs should provision their own variant administration in a fixture, and (b) which routing mode the app is actually in, so nav hrefs and deep links agree. Both are shillinq design calls, which is why I have not guessed at them.

rubenvdlinde:
The single largest cause of spec failure on a clean instance — and it isn't shillinq's code.

## What's happening

`firstrunwizard` is a **stock Nextcloud app**, enabled by default. On a fresh profile it renders a full-viewport opaque modal over the SPA:

```html
<div role="dialog" aria-modal="true" id="firstrunwizard"
     class="first-run-wizard modal-mask modal-mask--opaque">
```

Of the **90** `locator.click: Test timeout` failures in a full run, **84** logged `#firstrunwizard … subtree intercepts pointer events`.

⚠️ **The symptom does not look like an overlay.** Playwright reports the target as *"visible, enabled and stable"* and then retries the click until the test times out — so it reads as a dozen unrelated flaky-click defects rather than one cause. That is why it has survived this long.

33 spec files already carry a `dismissWizard()` helper for exactly this. But a helper that runs *inside* the test races the modal's own mount, and cannot help a spec whose first action is a click. Removing the app removes the race.

This is test-environment setup, not a product change: nothing in shillinq depends on `firstrunwizard`, and it is disabled only on the instance under test.

## Measured

Container serving current `development`, seeded by this same script, nothing else changed between runs:

| | passed | failed | skipped |
|---|---:|---:|---:|
| three worst spec files, wizard enabled | 0 | 26 | — |
| three worst spec files, wizard disabled | **17** | **9** | — |
| **full chromium suite, wizard enabled** | 200 | **71** | 3 |
| **full chromium suite, wizard disabled** | **213** | **49** | 7 |

Both full runs **completed** (274 tests, no `did not run`), so these are totals rather than the floor a truncated run reports.

## Why the earlier numbers on this repo were wrong

Every shillinq e2e figure I posted before today described a container mounting a worktree **101 commits behind `development`**, on which the API route `GET /bbv-dashboard` still shadowed the SPA page route of the same name and returned `412 CSRF check failed` on any deep link. I have retracted those on #564 and closed #618.

This measurement is from a **fresh container serving current `development`**, with both apps deployed at that ref and the register seeded by `ci-seed.sh`.

## What's left after this

49 failures remain, and they look like real work rather than environment:

- 5 are genuinely missing manifest routes, correctly caught: `/inventory/products`, `/inventory/product-attributes`, `/belastingen/vat-aangiften`, `/belastingen/vat-correcties`, `/belastingen/uitgestelde-belastingen/movements` — verified absent from all 590 declared page routes.
- the rest cluster in `bookings-resource-calendar`, `bank-statement-wizard`, `provincies-/waterschappen-bbv-variant`, `receipt-extraction-consume`.

I have not touched those — they need reading one at a time, not a sweep.

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -175,13 +175,34 @@ jobs:
       # these specs need for the same reason the Newman leg does: shillinq
       # delegates all object CRUD to /apps/openregister/api/objects.
       #
-      # ⚠️ `enable-playwright` / `playwright-test-path` are NOT repeated here.
-      # They are set once above (with the reasoning for the config-path choice),
-      # and a merge on 2026-08-16 re-introduced them at this point. YAML forbids
-      # duplicate keys in one mapping, so GitHub could not parse this file: it
-      # created a run, marked it `failure`, and ran ZERO jobs — which on every
-      # dashboard is indistinguishable from the red this repo already had.
-      # Set each input exactly once, in the block above.
+      # ⚠️ SET EXACTLY ONCE — here, and nowhere else in this file.
+      #
+      # These two keys have now been lost twice in one day, in opposite
+      # directions, by changes that could not see each other:
+      #
+      #   1. #564 and #440 each ADDED them to this block independently. YAML
+      #      forbids duplicate keys, so GitHub could not parse the file at all:
+      #      it created a run, marked it `failure`, and ran ZERO jobs. Fixed by
+      #      #856, which removed the copy further up.
+      #   2. #853 had branched before that fix, hit the same duplicate, and
+      #      resolved it by removing THIS copy instead. Net result: zero
+      #      declarations, `enable-playwright` fell back to its `false` default,
+      #      and `E2E Tests (Playwright)` reported **skipped** on development.
+      #
+      # Case 2 is the more dangerous one. A workflow that will not parse is at
+      # least loudly red; a skipped job renders in the Quality Report exactly
+      # like a passing one, so the 53 specs silently stopped running again with
+      # nothing on any dashboard to say so.
+      #
+      # `playwright-test-path` does double duty in the shared workflow: it is
+      # both the spec directory AND the first place the run step looks for a
+      # config (`${playwright-test-path}/playwright.config.ts`, falling back to
+      # the repo root). We ship tests/e2e/playwright.config.ts precisely so that
+      # lookup hits it — the run step passes no `--project`, so the ROOT config
+      # would additionally run `visual` (pixel baselines a CI runner cannot
+      # byte-match) and `docs-capture` (screenshot re-shoots with their own job).
+      enable-playwright: true
+      playwright-test-path: tests/e2e
       # Runs after `php -S` is up and before the first spec, with cwd set to the
       # Nextcloud server ROOT (hence the `apps/shillinq/` prefix).
       #
--- a/tests/e2e/ci-seed.sh
+++ b/tests/e2e/ci-seed.sh
@@ -166,6 +166,43 @@ else
 	echo "[ci-seed] no ./occ in $(pwd) — skipping the pretty-URL config step."
 fi
 
+# ── 0b. Disable Nextcloud's own first-run wizard ─────────────────────────────
+# `firstrunwizard` is a STOCK Nextcloud app, enabled by default, and on a fresh
+# profile it renders a full-viewport opaque modal over the SPA:
+#
+#     <div role="dialog" aria-modal="true" id="firstrunwizard"
+#          class="first-run-wizard modal-mask modal-mask--opaque">
+#
+# It is the single largest cause of spec failure on a clean instance. Measured
+# 2026-08-16 against a container serving current `development`, seeded by this
+# script: of 90 `locator.click: Test timeout` failures, **84** logged
+# `#firstrunwizard … subtree intercepts pointer events`.
+#
+# ⚠️ The symptom does NOT look like an overlay. Playwright reports the target as
+# "visible, enabled and stable" and then retries the click until the test times
+# out — so it presents as a dozen unrelated flaky-click defects, not one cause.
+#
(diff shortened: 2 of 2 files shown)


rubenvdlinde:
Added a second, separate defect to this PR: **`enable-playwright` is currently ABSENT on `development`**, so `E2E Tests (Playwright)` reports **skipped**.

It was lost twice in one day, in opposite directions, by changes that could not see each other:

1. #564 and #440 each **added** the pair to the same `with:` block → duplicate YAML keys → GitHub could not parse the file → a run with **zero jobs**. Fixed by #856 (mine), which removed the copy further up and left this one.
2. #853 had branched before that fix, hit the same duplicate, and resolved it by removing **this** copy instead. Between the two, both are gone.

⚠️ Case 2 is the worse failure. A workflow that will not parse is at least loudly red. A **skipped** job renders in the Quality Report exactly like a passing one — so the 53 specs stopped running again with nothing on any dashboard to say so. That is precisely the "a skipped job is not a pass" this block's original comment was written to prevent.

Verified after the restore: duplicate-key scan **NONE**; resolved inputs `enable-playwright=True`, `playwright-test-path=tests/e2e`, `playwright-seed-command` present.

I've also noted this on my own merge of #855 — its `E2E Tests (Playwright)` failure was not in the development baseline **because the job was skipped there**, not because it passed. A subset-vs-baseline rule cannot see that difference; mine compares `conclusion=="failure"` only. Worth treating `skipped` on the baseline as "no verdict" rather than "clean".

rubenvdlinde:
**CI result, with this PR applied: 242 passed / 25 failed / 7 skipped** of 274 (run 31948881650, E2E job 95169403461, 14.3m). The seed log confirms the new step ran:

```
[ci-seed] disabled Nextcloud's firstrunwizard (its modal intercepts pointer events).
Running 274 tests using 4 workers
  25 failed
  7 skipped
  242 passed (14.3m)
```

That is better than my local measurement (213/49/7) — CI installs fresh and `ci-seed.sh` lays down its full fixture set, so more specs find what they expect.

**Merging with the E2E job red, deliberately.** My own subset-vs-baseline rule held this PR, and it was right to by its own logic but wrong for this case:

- on `development` `E2E Tests (Playwright)` is **skipped**, because `enable-playwright` is currently absent (the second defect this PR fixes);
- a skipped job has **no verdict**, so nothing on the baseline can ever "cover" a real failure here;
- the rule compares `conclusion == "failure"` only, so it cannot tell "skipped" from "clean".

The choice is therefore between a job that reports **242/274 with 25 named failures**, and a job that reports **nothing at all while rendering like a pass**. The first is strictly better, and the second is the exact state this PR exists to end.

The 25 are attributed, not unknown: 9 are #862 (provincie BBV dashboard declares zero widgets), 2 are #860 (inventory product catalog ticked `[x]` but never built), 4 more are stale spec route strings fixed in #861, and the remainder cluster in bookings-resource-calendar / cashflow / ifrs15 and need reading one at a time.

Issue #871, opened 2026-08-16
49 lookups use findAll(['filters' => ['id' => ...]]), which matches nothing — 38 are lifecycle guards


rubenvdlinde:
## Summary

49 live call sites query OpenRegister with `findAll(['filters' => ['id' => $someId]])`. That shape matches **zero rows for every value**, silently. 38 of them are lifecycle guards, where it means the guard's object-state check can never succeed and every call takes the not-found branch.

One instance of this is fixed in #870 (`FiscalYearContextService`, where it silently broke REQ-BBVW-006's fiscal-year inheritance). This issue covers the rest.

## The mechanism

`filters` addresses the object's **JSON properties**. The ObjectEntity's `id` is the entity's own column; it is merged into the serialised output afterwards, which is what makes the shape look plausible — a record fetched over the API visibly *has* an `id` key.

Measured in-container against real data, filtering each schema by its own real id:

```
Receipt            total=1    filters.id(4c2c0d82…)=0
Administration     total=1    filters.id(1c13b57b…)=0
BudgetBBVMapping   total=1    filters.id(d8cb5294…)=0
```

`findAll([])` returns the row; `findAll(['filters' => ['id' => <that row's id>]])` returns nothing. No exception, no log.

The trap that hides it: the **HTTP list API does answer `?id=…`** — it maps that onto the entity column — so the same lookup looks healthy when probed over the wire and returns nothing in PHP.

Note `lib/Util/ObjectIdentifier.php`'s docblock already describes a related failure for this shape (`SQLSTATE[22P02] invalid input syntax for type bigint`). On this rig it does not throw, it returns empty — which is the worse of the two, because a throw would at least be loud.

## Why it is invisible

Every affected guard handles "not found" gracefully. `GLReversalGuard::isReversed()` is representative:

```php
$found = $objectService->setRegister(...)->setSchema('GLTransaction')
    ->findAll(['filters' => ['id' => $glTransactionId], 'limit' => 1]);

if (is_array($found) === false || $found === []) {
    $this->logger->info('GLReversalGuard: linked GLTransaction not found — denying void', ...);
    return false;
}
```

That reads as a careful fail-closed path. In practice it is the *only* path — the state check never runs, so the guard's actual rule is dead code. Where a guard fails closed this blocks legitimate operations; where one fails open the consequence is worse. Each site needs its own read.

## Inventory

**Lifecycle guards — 38 sites across 34 files:**
[...]

Issue #866, opened 2026-08-16
The Provincies BBV dashboard is declared in a manifest vocabulary no renderer implements — the page is empty for every user


rubenvdlinde:
`src/manifest.d/bookkeeping-provincies-bbv-variant.json` declares the whole body of `/bbv-provincie/compliance-dashboard` under

```json
"config": { "register": "shillinq", "schema": "Budget",
            "dashboard": { "kpis": [...], "charts": [...], "exceptions": {...}, "filters": [...] } }
```

**Nothing reads `config.dashboard`.** The page mounts an empty `CnDashboardPage`: no KPI, no chart, no exceptions block, no filter facet, for every visitor.

## Measured, not assumed

- `CnDashboardPage` declares exactly two content props — `widgets` and `layout` (plus `dateRange` / `headerActions`). There is no `kpis`, `charts`, `exceptions` or `filters` prop.
- `CnPageRenderer` has **zero** references to `config.dashboard`.
- `grep -rn kpis` across all of `@conduction/nextcloud-vue/src` returns **two** hits, both inside a docblock *example* naming a widget id.
- Of the **twelve** `type:"dashboard"` pages shillinq declares, **eleven** use `widgets` + `layout`. This page is the **only** `config.dashboard` in the entire manifest — a one-off written against a schema that does not exist.

The same fragment's linker index declares `config.bulkActions[]` and `config.mappingStatus`, which are likewise not CnIndexPage props (zero references). Its detail page declares `config.fields[]` and `config.indexRoute`, also unread.

## Why this is not a mechanical translation

Two of the four KPIs (`committed`, `spent`) and **both** chart series are sourced from the named OpenRegister aggregation `programmeBudgetVsActuals`. `CnStatsBlockWidget` and `CnChartWidget` fetch `/apps/openregister/api/objects/aggregations/{register}/{schema}/value` with a `metric` + `field` — they cannot address a named schema aggregation. Converting the fragment therefore needs a **product decision** about how programme budget-vs-actuals is surfaced, so it is escalated rather than guessed. Only `total-budget` and `remaining` (plain `sum` over `Budget.totalAmount`) map cleanly today, and shipping 2 of 4 KPIs and no charts would be worse than the honest red.

## E2E status

[...]

Issue #865, opened 2026-08-16
REQ-CF-016 (cashflow PDF export) is unimplemented end to end: renderer with no caller, button with no host


rubenvdlinde:
`openspec/specs/bookkeeping-cashflow-13wk/spec.md` still carries **REQ-CF-016: PDF export for bank/accountant meetings** as a current requirement, with a scenario that ends *"WHEN 'Export PDF' button clicked THEN system generates PDF with all sections above"*.

**It is unimplemented end to end.**

## Measured on `development`

| layer | state |
|---|---|
| renderer | `lib/Service/CashflowPdfRenderer.php` exists; `render()` has **zero callers** anywhere in `lib/` |
| route | `appinfo/routes.php` contains **no** cashflow export route |
| manifest | `/cashflow/dashboard` (`src/manifest.json`) declares `widgets` + `layout` + `documentationUrl` and **no `headerActions`** — so no export affordance renders |
| component | the only "Export PDF" button in the codebase is in `src/components/cashflow/CashflowDashboard.vue`, which is an **orphan**: no import, no `src/registry.js` entry, no manifest page binds it as `component`. Even if it mounted, its click only `$emit('export-pdf', …)` and nothing listens |

So there is a renderer with no caller, a button with no host, and a host page with no action.

## How it surfaced

`tests/e2e/cashflow-13wk.spec.ts` → *"dashboard exposes an Export PDF affordance"* fails with `getByRole('button', { name: /export.*pdf/i })` → element(s) not found.

The test has been **left asserting on purpose**. Quarantining it, or relaxing it to "some button exists", would convert a missing reporting capability into an invisible one.

## Ask

Either build REQ-CF-016 (route + controller calling `CashflowPdfRenderer::render()`, plus a `headerActions[]` entry on the dashboard page), or withdraw REQ-CF-016 from the spec and delete the orphan `CashflowDashboard.vue` skeleton. A green E2E column must not be the third option.

⚠️ Related shape worth a wider sweep: **an orphan Vue component whose testids leak into a spec file**. The same E2E file asserted `data-widget="cashflow-13week-chart"`, an attribute that exists ONLY inside that orphan component — the author read the component, not the shipped page. `@conduction/nextcloud-vue` emits no `data-widget` attribute at all, and its charts are ApexCharts **SVG**, never `<canvas>`.

Issue #860, opened 2026-08-16
Inventory product catalog: Task 14 ticked [x], but neither the manifest pages nor the Product schemas exist


rubenvdlinde:
`openspec/changes/archive/2026-06-14-inventory-product-catalog/tasks.md` Task 14 is ticked `[x]`:

> Added Products navigation to `src/manifest.json`: new "Inventory" top-level menu (order 22) with Products (index, route `/inventory/products`, schema `Product`, columns: sku/name/category/unitPrice/status) and ProductAttributes (index, route `/inventory/product-attributes`, schema `ProductAttribute`). Detail pages for both. All rendered by `CnIndexPage`/`CnDetailPage` per REQ-IPC-008.

**None of it exists.**

## Measured

Manifest — all 590 declared page routes, across `src/manifest.json` and every `src/manifest.d/*.json`:

```
pages whose id / route / title mentions "product": 0
```

The 34 `/inventory/*` routes that DO exist are barcodes, count-templates, cycle-counts, expiry-alerts, lots, low-stock-alerts, mobile/*, posting-config, posting-history, reorder-rules, reserve-stock, reserved-stock, stock*, valuation*, variance-reports. No `products`, no `product-attributes`.

Schemas, on a live instance running current `development`:

```
GET /apps/openregister/api/objects/shillinq/Product          {"message":"Schema not found: 'Product'"}
GET /apps/openregister/api/objects/shillinq/ProductAttribute {"message":"Schema not found: 'ProductAttribute'"}
```

So the pages are absent, and the schemas they would bind to are absent too. This is not a wiring slip — the capability was never built.

The only reference to either route anywhere outside `tests/` is the archived tasks.md line above.

## How it surfaced

`tests/e2e/spec-coverage/inventory.spec.ts` asserts both routes and fails with:

```
route /inventory/products must be declared by the manifest and resolve
route /inventory/product-attributes must be declared by the manifest and resolve
```

The spec is right and the tick is wrong. I have left both entries asserting rather than quarantining them — a spec that fails because a capability is missing is the spec working. Silencing it would convert a known gap into an invisible one.

## Ask

Either build the capability (schemas + manifest pages, as the archived task describes) or untick Task 14 and record that it was descoped. A `[x]` on an archived change is the only record anyone will find, and right now it asserts something false.

[...]

PR #882, opened 2026-08-17
fix(ci): clear four of the six Hydra gates red on development, and the Newman filter that matched nothing

rubenvdlinde:
`development`'s **push** run is judged full-repo — `on: push` has no diff base, so every Hydra gate scans the entire tree and the `Hydra Gates` cell stays red until the whole backlog is cleared, not just what a PR touches. Six gates were failing. This clears four of them, plus the single Newman assertion.

| gate | before | after |
|---|---|---|
| gate-25 contract-coverage | FAIL — 51 | **PASS** |
| gate-46 spec-anchor-existence | FAIL — 1 | **PASS** |
| gate-54 relation-dialect | FAIL — 85 | **PASS** |
| gate-57 orphaned-write-capability | FAIL — 1 | **PASS** |
| gate-26 visual-coverage | FAIL — 2 | FAIL — 1 |
| gate-7 no-admin-idor | FAIL — 4 | FAIL — 4 |
| `Integration Tests (Newman)` | 4/5 assertions | **5/5, verified live** |

Measured with the gate suite from `ConductionNL/.github@main`, not a vendored copy, before and after. `gate-22` and `gate-53` fail in my local runs only — both report "ajv is not resolvable", an absent `node_modules`, and both are green in CI.

## Newman — the request was wrong, not the assertion

One assertion has been red: `REQ-008 — Mijn Contracten filter returns the tenderned-sourced row`, `expected +0 to be at least 1`. The two POSTs before it return 201 and pass their own assertions, so the row exists; the filtered GET answers **200 with an empty array**.

OpenRegister's `ObjectsController::getConfig()` assigns the **whole** request parameter bag to `filters`, with reserved parameters underscore-prefixed (`_limit`, `_page`, `_search`, …). So `filters[source]=tenderned` arrives as a filter on a property literally named `filters`. No schema declares one, and **a filter on a non-property matches nothing silently** — it does not 400.

Measured live against openregister with one seeded Verplichting, rather than reasoned about:

```
?filters[source]=tenderned&filters[sourceReference]=TN-PROBE-0001  -> 200, total 0
?source=tenderned&sourceReference=TN-PROBE-0001                    -> 200, total 1
```

and the collection then runs **5/5 green end to end** against a live instance. No assertion was relaxed, retargeted or deleted — the assertion was correct throughout and is what caught this. The request now carries a description recording the mechanism so the `filters[…]` form is not reintroduced.

## gate-25 — 51 real contract tests, zero excludes

24 controller test classes (22 new, 2 extended), **310 tests, all green**, each constructing the controller with mocked collaborators and asserting observable behaviour. **No `@contract exclude` was written for any endpoint** — a gate that reads prose can be switched off by apologising to it, and 51 apologies would have measured nothing.

⚠️ **Honest caveat about the gate, not the tests:** gate-25's PHPUnit arm searches the concatenated text of every `*Test.php` for a **bare method name**, with no association to the controller under test. A green gate-25 is therefore weak evidence. The tests here are real regardless — they were written against the controllers, not against the matcher — but do not read the PASS as proof of endpoint coverage. A `.github` fix is being tracked separately.

Two live defects surfaced by writing them, recorded but **not** fixed here (they are `lib/` changes that deserve their own PR):
- `Content-Disposition` filenames are silently discarded on `IcpController::renderInvoicePdf` and `InvoiceApiController::pdf` — `DataDisplayResponse::__construct()` re-adds `inline; filename=""` *after* applying the headers passed to it, so the client always receives an empty filename. The tests assert the observed value with a comment naming the cause, plus a companion test proving a hostile filename still cannot inject a header.
- `DBAController::jsonBody()` reads `php://input` directly, so `IRequest` cannot feed it.

## gate-54 — 84 retired relation blocks, judged individually

ADR-062 rule 7 retired the per-schema `x-openregister-relations` block in favour of a property-level `$ref`. 142 relation entries across 35 register fragments were read one at a time: **58 migrated** to a property `$ref`, **84 deleted without replacement**.

Deletion is safe because **nothing consumes the block** — verified, not assumed: `Schema.php:2303` is a persist allow-list, and `AnnotationNotificationDispatcher`'s `relation` recipient kind resolves the *notification's own* `recipients[].relation` against object data.

The 84 fall into three groups, each with a reason a `$ref` would be wrong:
- inverse / one-to-many sides, where the FK lives on the *other* schema;
- `relatedField` names an **operator-assigned business key**, not the object identity — `Service.serviceId` is documented as "Operator-assigned unique service identifier", example `svc-001`; likewise `Verplichting.commitmentNumber`, `Account.accountNumber`, `TenderNedAanbesteding.tenderId` and others. A `$ref` there names a target OpenRegister cannot reach through that value;
- targets outside the register set (`Person`, `pipelinq:Product`, …).

**`format: uuid` was deliberately withheld everywhere.** A scalar bare-slug `$ref` is `unset()` before validation (`ValidateObject.php:603-608`), so it is inert; `format: uuid` **is** enforced and would reject shipped data — `integral-cost-price-example-q1-2026.json` writes `"commercialActivityId": "MO-SP-015"` into a `relatedField: id` FK. The same finding demoted one migration candidate: an `items.$ref` is *not* stripped (`ValidateObject.php:574-587` replaces it with a UUID pattern), so `AlgemeenBelangBesluit.concernsActivities`, whose seed holds `["MO-SP-016"]`, would have failed on save.

The one `(case-exact)` finding: `Contract.decisions` `$ref`'d decidesk's `Decision`, declaring `x-external-register` on `items` while the gate reads it from the **property**. Fixed as the gate prescribes.

## gate-57 — a dispatch that could never be dispatched

`DunningRunService::transferToIncasso()` performs the REQ-CCD-008 hand-off to the bound collection-agency adapter and seals the `DunningRun` on a DELIVERED outcome. Its only callers were two unit tests. `dossier()` composed the stage-5 bundle and handed it back to the client; **nothing then sent it.** Task-20's transfer half was implemented, unit-tested and ticked done while being unreachable at runtime.

Wired rather than deleted, since the composer half is already live and routed. `POST /api/dunning/incasso/transfer` mirrors the sibling `dossier()` endpoint exactly, including `AdministrationContextService::canAccess()` with a **masked 404 rather than a 403**, so the response cannot become an existence oracle for another tenant's ids.

The dossier is composed **server-side**, never accepted from the request body — it is the evidence bundle a debt-collection agency acts on, so a caller must not be able to name the debt. `testTransferComposesServerSideAndReturnsOutcome` is a positive control, not a tick: pointing the controller at the request's own `dossier` param makes it report

```
The dispatched dossier must be the composed one, never the request body
-    'invoice' => 'inv-1'
+    'forged' => true
```

The recording double uses `willReturnCallback` in the service's own parameter order rather than `->with()` — a PHPUnit mock resolves named arguments against its own signature and then invokes the callback **positionally**.

## gate-46 and half of gate-26

`provincies-bbv-routes-smoke.spec.ts` pointed at `#smoke-tests`, a fragment naming nothing; the task it implements is `Task 26: Smoke tests`, whose own DONE note names this exact file. Verified by running the helper in both directions — it writes findings to a **log** and always exits 0, so its exit code proves nothing: `#task-26` → empty, `#task-999` → reported.

`src/views/settings/UserSettings.vue` is deleted. It has existed since the initial commit, has never had an importer (`git log -S`), appears in neither the registry nor the manifest, and its body is an `NcAppSettingsDialog` whose only content reads "User settings will appear here in a future update." A visual baseline for a component nothing can render measures nothing.

## What this PR does NOT fix

- **gate-26's remaining finding**, `StandardsPolicyEditor.vue`, is a real page (`/settings/accounting-standards`). It needs a test that actually drives it, and the dev container's bundle predates `development`, so nothing runnable locally would be evidence about the code under test.
- **gate-7's 4 IDOR findings** are being handled separately.
- **`E2E Tests (Playwright)`** is red on 8 tests that a previous crew left asserting **on purpose** — four unimplemented capabilities (the BBV compliance dashboard's `config.dashboard` fragment that no renderer reads, `config.filters[]` likewise, the cashflow PDF export, and the inventory product catalog whose Task 14 is ticked but never shipped). Those are product changes or spec withdrawals, not test fixes, and turning them green here would convert four missing capabilities into invisible ones.


rubenvdlinde:
Reviewed from a separate session that was independently re-measuring `development`'s four failing jobs. I reached the same conclusion as this PR on all four gate-7 findings (two stateless calculators, `narrative()` never using its path param, and spend-analytics being the only real one) before reading the exemptions here — so those corroborate rather than duplicate.

One blocking finding, on `c179475a`:

### `transferToIncasso()` is wired, but its sealing lookup is dead against real OpenRegister

`DunningController::transfer()` correctly closes gate-57 — the diagnosis is right, the dossier really was composed and never sent. But the method it now makes reachable still contains, at `lib/Service/DunningRunService.php:951`:

```php
$run = $this->fetchOne(schema: 'DunningRun', filters: ['id' => $dunningRunId]);
if ($run === null) { $this->logger->warning(…); return $result; }
```

`fetchOne()` (`:1078`) is `findAll(['filters' => $filters])`. OpenRegister's `filters` addresses the object's **JSON properties**; `id` is its own column, merged into the serialised output afterwards — so this matches nothing, silently, on every call. `$run === null` always.

The order of operations makes the consequence specific: the adapter call is *before* the lookup, so after this PR

- the dossier **is** dispatched to the collection agency;
- the linked `DunningRun` is **never** sealed to `lifecycleState=locked` — REQ-CCD-002 immutability and the REQ-CCD-008 lock are not written;
- the provider's `dossierId` is **never** stamped, so the evidence trail this method exists to create is lost;
- and because the run stays `executed` rather than `locked`, nothing prevents the same stage-5 dossier being dispatched to the agency **again**.

The same shape is at `:287`, `:506` and `:1007` in this file.

### Why the tests are green over it

`DunningRunServiceTest` uses `tests/Unit/Service/InMemoryObjectService.php`, whose `findAll()` (`:106-128`) matches filters against **plain PHP array keys, `id` included**:

```php
$actual = $row[$key] ?? null;                       // :118
if ((string)$actual !== (string)$expected) { return false; }
```

So the double answers a lookup real OpenRegister refuses, and `testTransferToIncassoLocksRunOnDelivery` asserts `lifecycleState === 'locked'` against a store that can produce it while production cannot. Same class as the six doubles already recorded on the coordination board; this is a seventh, still live.

### Suggestion

Fix the `:951` lookup in this PR — a by-id read rather than a property filter — or hold the routing commit until it is fixed. Wiring an orphaned write capability does not make it work; it makes its latent defects live, and gate-57 says nothing about whether the path it unblocks is correct.

---

Two other things from the same measurement, neither blocking:

- **The Newman change (`a427c507`) is right, and it corrects the coordination board.** I re-measured independently against a live rig with a positive *and* a negative control: `filters[source]=manual` → total 0, `source=manual` → total 3, `source=NOSUCHVALUE` → total 0, unfiltered → total 4. The board had this recorded twice as an OpenRegister defect; it was caller error. `ObjectsController::getConfig()` does `'filters' => $params` over the whole bag. Worth noting the same function does `unset($params['id'])`, so `?id=<uuid>` on the list endpoint returns the **entire collection** (measured: total 4 of 4, and `?id=NOSUCH` likewise) rather than nothing — a worse trap than the one the board describes.

- **This PR's E2E is 9 where `development` is 8, and the 9th is a flake, not a regression.** `waterschappen-bbv-variant.spec.ts:170` passed at `614acbea` (12.5s) and failed at `f2c6883e`; `git diff --stat` between those two heads is `ExpenseClaimGuard.php` plus thirteen PHPUnit files — zero `src/`, zero manifest, zero register fragments, so nothing in it can reach that button. The cause is `click()` → `waitForLoadState('domcontentloaded')` → `page.url()`: an SPA router push inside an already-loaded document, so the wait resolves in the same tick. Fixed in #883 (test-only, one file) with `expect.poll` at that site and its unfired twin at `:117` — identical regexes, untouched locators. #883 does not touch any file in this PR.

rubenvdlinde:
### Correction to my own comment above — the 9th E2E failure is NOT a flake

I said `waterschappen-bbv-variant.spec.ts:170` was non-deterministic, on the
grounds that `git diff --stat 614acbea f2c6883e` is `ExpenseClaimGuard.php` plus
thirteen PHPUnit files and therefore cannot reach that button. **I then re-ran
the E2E job at the identical sha to check, and it refuted me.**

Attempt 2 of run `31995390615` (job `95294123132`, log asserted at 347,004 bytes)
failed the **same 9** tests, with a **byte-identical** message:

```
Expected pattern: /budget-mappings\/new|budget-mappings\/[^/]+$/
Received string:  "http://localhost:8080/apps/shillinq/budget-mappings"
```

So the picture is:

| tree | runs | this test |
|---|---|---|
| `development` @ `5914bf34` (#880+#881, no #882) | 1 | ✓ |
| this PR @ `614acbea` (#882 changes, **no #881**) | 1 | ✓ |
| this PR @ `f2c6883e` (#882 changes **+ #881**) | **2** | ✗ ✗ |

**It reproduces, and it is red only on the combination.** The diff that
introduces the combination genuinely cannot touch a Vue button, so the cause is
an interaction, an execution-order effect or a seed difference — none of which a
file-level diff shows, and all of which a 273-test suite against one seeded
Nextcloud instance is full of. I could not attribute it further.

My mistake is worth naming because it is cheap to repeat: **"the diff cannot
touch the subject" bounds what a change directly modifies, and nothing else. It
is not evidence of non-determinism, and using it to skip a rerun is a check that
did not run.**

Practical consequence for this PR: under the fleet's parity rule the failing
*job* set here is still a subset of `development`'s, so it is mergeable on the
letter of it. But a test that is green on `development` and reproducibly red
here is precisely the shape that rule exists to catch. **Worth bisecting against
#881 before merging, or merging in the knowledge that it is unexplained.**

Also correcting what I said about #883: it repairs the instrument at that site
(`waitForLoadState` after an SPA router push is not a wait) but it is **not shown
to fix this failure** — it is not on this branch, and the test already passed on
`development`. It cannot mask it either: with polling, a CTA that never navigates
still fails, 15 s later. #883 is now described that way.

The `transferToIncasso()` finding in my first comment is unaffected and still
stands.

rubenvdlinde:
### The gate-57 closure carried a live defect — fixed on this branch at `5df5ac61`

I verified the report independently before changing anything, and it is correct.

In `lib/Service/DunningRunService.php`, `transferToIncasso()` called the collection-agency adapter **first** (`:933`) and only then looked the run up (`:951`) with `fetchOne(schema: 'DunningRun', filters: ['id' => $dunningRunId])`. `filters` addresses the object's JSON properties; `id` is the ObjectEntity's own column, so that shape matches **zero rows against real OpenRegister for every value**, silently. Wiring the method into production via `DunningController::transfer()` therefore shipped: the dossier **is** dispatched to the agency, the `DunningRun` is **never** sealed to `locked`, `dossierId` is **never** stamped, and **nothing prevents re-dispatching the same dossier**.

The two existing unit tests could not see it — `InMemoryObjectService::findAll()` matches `id` as a plain PHP array key, so the double answers a query the engine refuses.

**What changed**

- `fetchById()` — identity lookup through `ObjectIdentifier::findOne()`. All **four** `filters['id']` sites in this service moved onto it; the ladder, pause and PostNL-evidence lookups were dead against real OR too, so `resolveLadderForKlant()` / `tickInvoice()` / `resumePause()` took their not-found branch on every production call.
- `transferToIncasso()` reordered so a failed seal cannot leave a dispatched dossier: resolve the run (absent ⇒ refuse, **no dispatch**) → refuse an already-`locked` run → **write the seal** → dispatch → stamp `dossierId`. A non-DELIVERED outcome releases the seal back to `executed`; if that release fails the run stays locked, i.e. **fails closed** — a blocked retry is recoverable, a duplicate dossier at a collection agency is not.

**Positive control, both directions** (same tree, `phpunit-unit.xml`):

| | result |
|---|---|
| faithful double + **shipped** lib | **5 errors + 5 failures** of 25 |
| faithful double + **fixed** lib | **25 passed, 70 assertions** |

The three new tests each fail on the shipped code for the right reason: a spy adapter reads the store from *inside* `transfer()` and sees `executed`; the absent-run and sealed-run cases both record the adapter being called (`1 !== 0`).

`tests/Unit/Service/Support/OpenRegisterFaithfulObjectService.php` is the in-memory double with the engine's identifier semantics restored. It is **not** made the default: flipping the lenient one red-lines **27 tests across six services** carrying the same defect (~60 live `filters['id']` sites, #871), which is a product decision rather than a test-side one. The lenient double's docblock now says so and names the faithful subclass.

Full unit suite: **4446 tests / 44851 assertions / 0 failures / 0 errors** (branch baseline 4443 / 44844 / 0 / 0). phpcs, phpmd and psalm clean; phpstan cannot run locally.

rubenvdlinde:
### Two follow-ups on the same commit series (head `848c5e4f`)

- `52a6c4c0` — **`RetainerGuard` and `ProvisionGuard` carry the byte-identical dead `resolveObject()`**, and the consequences are opposite, so each caller had to be read: `RetainerGuard::canMaterializeDrawdown()` **fails OPEN** (its rate-immutability comparison sits behind `if ($pool !== null …)`, a narrow cross-app concession that widened to 100% of calls), while `ProvisionGuard::canActivateProvision()` **fails CLOSED permanently** — every `restructuring` and every `claims` provision was blocked from activating, and that guard is live via the `requires:` clause in `bookkeeping-voorzieningen-claims.json`. Both repaired narrowly; **no rule relaxed or added**, the IAS 37 §72/§75, REQ-PROV-006 and REQ-RETN-002 checks are unchanged and merely evaluable now. The other ~58 `filters['id']` sites in `lib/` are untouched — that is #871's sweep and out of scope here.

  Positive control: unfixed `RetainerGuard` → 1 failure (`true is false`, the fail-open); unfixed `ProvisionGuard` → 2 failures. Fixed: 11/11 and 21/21. The three **negative** controls (matching rate materialises, late plan still blocks, missing memo still blocks) pass in **both** directions, so the repair evaluates the rules rather than deleting them.

  ⚠️ **`RetainerGuard` has no production caller at all** — no controller, no `requires:` clause. Its repair is latent until someone wires it; `RetainerBillingFragmentTest`'s docblock claim that it is "invoked by the controller layer" is not true today. Reported, not wired.

- `848c5e4f` — `resolveIncassoAdapter()` now runs **before** the seal is written. It throws when the binding is missing, and behind the seal that would leave a run permanently `locked` over a configuration error with nothing dispatched. Fail-closed is the right direction for a failed dispatch and the wrong one for a failed container lookup.

Full unit suite on the final head: **4451 tests / 44856 assertions / 0 failures / 0 errors / 1 skipped**. phpcs (scoped to `lib`), phpmd (both rulesets) and psalm exit 0.

⚠️ Two caveats on my local numbers, so nobody over-reads them: they were measured on **PHP 8.2** with a `vendor/` copied from another worktree, and phpstan cannot run locally here (`phpstan.neon` points at an uninstalled `nextcloud/ocp` path) — CI is the authority on both. Also: a **symlinked** `vendor/` initially resolved `OCA\Shillinq\` to the *other* worktree's `lib/`, which would have made every edit invisible while the tests still ran and reported green; I asserted `ReflectionClass(DunningRunService)->getFileName()` inside my own tree before trusting any run, and re-measured everything after.

E2E is expected to stay red on its 8 product gaps (#865, #866/#862, #860) — nothing here targets them, and they should not be made to pass.

rubenvdlinde:
### CI on `848c5e4f`: **41 pass, 1 fail (E2E), 1 pending** — and the E2E red is NOT this branch's

**Hydra Gates: PASS.** All six gates stay closed with these commits. phpstan, phpcs, phpmd, psalm, Newman, Coverage Baseline Protection, Frontend Build/Checks, CodeQL: pass. Five of six PHPUnit cells green (8.3/32, 8.3/33, 8.4/32, 8.4/33, 8.4/34); 8.3/stable34 still running.

E2E is red, as expected — but the counts needed checking rather than trusting, because **identical counts are not identical content**:

| run | E2E result |
|---|---|
| `development` (baseline) | **8 failed** / 258 passed |
| this branch **before** my commits (`f2c6883e`) | **9 failed** / 258 passed |
| this branch **at `848c5e4f`** | **9 failed** / 258 passed |

Same number on the branch before and after, but **not the same tests**. Diffing the failing titles:

- **new at my head:** `pipelinq-customer-bridge.spec.ts:142 › test-connection button is present and clickable` — a `locator.click` **60 s timeout**, and it *passed* on `f2c6883e`.
- **gone at my head:** `waterschappen-bbv-variant.spec.ts › Add CTA navigates to the new-mapping detail page` — it *failed* on `f2c6883e` and passes here.

A one-for-one swap of two timeout-shaped frontend specs, each passing in the other run. The **8 canonical product gaps are byte-identical across all three runs** (cashflow Export-PDF affordance, 5× provincies-bbv KPI/chart/facet/empty-state, 2× inventory route resolution — #865, #866/#862, #860) and remain red, which is the intended state.

My diff is **8 files, all PHP** (`lib/Service/DunningRunService.php`, `lib/Lifecycle/{Retainer,Provision}Guard.php`, five test files) — no Vue, no manifest, no route, nothing the pipelinq admin panel or the BBV mapping index can reach. So the 9th failure is branch flake, not a regression from these commits.

PR #901, opened 2026-08-17
feat(cashflow): build the REQ-CF-016 PDF export end to end (#865)

### LINKED ISSUE #865: REQ-CF-016 (cashflow PDF export) is unimplemented end to end: renderer with no caller, button with no host

rubenvdlinde:
Closes #865.

`CashflowPdfRenderer.php` existed with **zero callers**, `appinfo/routes.php`
registered **no** cashflow export route, and the only "Export PDF" button in the
codebase lived in an **orphan component** that never mounts and whose click
`$emit`s into nothing. REQ-CF-016 was written and unreachable. This builds it.

`tests/e2e/cashflow-13wk.spec.ts:148` was red **on purpose** and is now green by
being **satisfied**. The spec file is byte-identical to `development` — no
retarget, no relax, no skip.

---

## What is wired

| layer | file |
|---|---|
| PDF bytes | `lib/Util/PdfDocument.php` (new) — dependency-free object serialiser + xref/trailer, lifted from `InvoicePdfGenerator`'s existing byte-writer rather than copied a second time |
| document | `CashflowPdfRenderer::renderPdf()` (new method) — the same report as a real `%PDF-1.7` binary, plus the REQ-CF-016 §2 closing-balance bar chart as vector rectangles coloured by `bufferStatus`. `render()` is byte-identical. |
| data | `lib/Service/CashflowExportService.php` (new) — horizon + weeks + top-5 AR customers + recurring costs out of OpenRegister, scoped to the caller's `AdministrationMembership` set |
| route | `lib/Controller/CashflowExportController.php` (new) + `POST /api/cashflow/export-pdf` |
| affordance | `config.headerActions[]` on the `CashflowDashboard` page in `src/manifest.json` |

### Why `render()` was not simply given a `.pdf` name

`render()` returns `text/plain` and a `.txt` filename. Advertising
`application/pdf` over those bytes would produce a file that opens in **no**
viewer while every route, status and "a button exists" assertion stayed green —
which is the exact shape #865 was filed about. `renderPdf()` is why the
`application/pdf` assertion in the new tests means something.

### Why there is no component to register

Board §8 records softwarecatalog/procest shipping a page declared in the
manifest whose component was missing from `customComponents.js`, rendering
*"This page is empty"* while its backend route was live. That shape is
**structurally impossible here**: `CnDashboardPage` passes `config.headerActions[]`
straight to `CnActionButtons`, which renders each entry as a plain **visible**
`NcButton` — not the collapsed `⋯` overflow `CnActionsBar` uses on an *index*
page, which is the trap already recorded in
`src/manifest.d/recurring-invoicing.json`. No custom component participates, so
there is no registration that can be forgotten.

The orphan `src/components/cashflow/CashflowDashboard.vue` is **left in place**
(its crisis-banner and scenario-switcher skeletons are unspent design) but now
says at the top of the file that it is not the export path. Deleting it is a
separate call.

### No IDOR surface, by construction

`exportPdf()` takes **no request parameters**. The horizon is resolved
server-side from `AdministrationContextService::accessibleAdministrationIds()`
(REQ-MA-001). Nothing caller-supplied crosses the boundary, so no per-object
guard appears in the method body — the absence is the design, and the docblock
says so. `#[NoAdminRequired]`, no `#[NoCSRFRequired]`: the SPA POSTs through
`@nextcloud/axios`, which carries the request token.

Every OpenRegister lookup filters on a declared schema property
(`administrationId` / `horizonId`), never on `id`.

---

## Proof the PDF is genuinely produced and downloadable

Not "a route returned 200". Measured on the rig (`sqdev-nc`, :8119), with the
changed `lib/` `docker cp`'d file-by-file and **md5-verified against the served
copy** (the container is a Docker volume, not a bind mount), the app warmed
afterwards, and the built bundle `docker cp`'d with
`md5 js/shillinq-main.js == md5 <served>` = `d7abe16b20c1b2d2c2d6c774a2869cb9`.

**Live red → green on the tenant guard, which is the same call twice:**

```
POST /apps/shillinq/api/cashflow/export-pdf
  -> http=404  {"error":"no_cashflow_horizon"}
```

The rig's seeded horizon belongs to `adm-consultancy-nl`; `admin` held a
membership only for `ADM-001`. The guard was correctly refusing. Granting
`admin` one membership for `adm-consultancy-nl` and repeating the identical call:

```
http=200  type=application/pdf  size=2629
Content-Disposition: attachment; filename=cashflow-hor-2026-w22-erik-2026-08-17.pdf
Content-Type: application/pdf
%PDF-1.7 \n % 342 343 317 323 \n 1 0 obj \n << /Type /Catalog
```

The membership was **deleted again** afterwards and the endpoint re-measured
back to `404` — the rig is as it was found.

**Parsed by an independent library**, not by the code that wrote it
(`pypdf.PdfReader`, 2 pages):

```
page 1: 13-WEEK CLOSING BALANCE
        Horizon 2026-05-25 .. 2026-08-23   blue = above buffer, ...
page 2: 13-WEEK CASHFLOW FORECAST
        Administration: adm-consultancy-nl
        Week   Inflows      Outflows     Net          Eind Saldo   Buffer
        22     12500.00     7225.00      5275.00      77775.00     ABOVE_BUFFER
        TOP CUSTOMERS (BETALINGSGEDRAG)
        klant-acme-bv                  avg offset +13 days, confidence 0.75
        RECURRING COSTS
        Huur kantoor                   MONTHLY      1200.00 EUR (FIXED)
        BAV-verzekering                ANNUALLY     620.00 EUR (CPI_PAST_YEAR)
        METHODOLOGY ...
```

On a 13-week fixture the chart content stream parses to **13 filled rectangles**
with distinct geometry plus the zero line. `file(1)` reports
`PDF document, version 1.7`.

⚠️ Board §6 warned a download path might not even be constructible locally
because `symfony/http-foundation` is absent from shillinq's vendor. **That was
the stale shared vendor, not the lock** — see the findings section.

---

## RED-then-green controls

**Unit, measured one file at a time** because a missing class aborts PHPUnit
collection for the whole run:

| file | without the wiring | with it |
|---|---|---|
| `CashflowPdfRendererTest` | `Tests: 12, Errors: 7` — and the **5 pre-existing `render()` tests still passed**, the control that the file itself is healthy | 13 / 13 |
| `CashflowExportServiceTest` | `Tests: 8, Assertions: 0, Errors: 8` — `Class "…\CashflowExportService" not found` | 8 / 8 |
| `CashflowExportControllerTest` | collection abort, class not found | 4 / 4 |

**No mock observes an argument anywhere in the new tests.** A PHPUnit mock
cannot see a named argument — it resolves the call against its own signature
then invokes the return callback positionally — so an expectation over this
app's named-argument style would pin the double's defaults. The controller's
export-service double is **hand-written**; the export service runs against the
repo's `InMemoryObjectServiceStub`, which really filters. **Tenant scoping is
proved by seeding a SECOND administration and asserting its data is absent from
the produced bytes**, not by asserting a query shape.

Two assertions carry their own negative half, because the positive half alone
would pass on a decoration:
- the chart is asserted **present with weeks and absent without them**;
- a `BELOW_BUFFER` week is asserted **red and below the zero line**, so three
  weeks at −50k / 0 / +50k cannot all render identically upward.

**e2e, on the rig, `PLAYWRIGHT_BASE_URL=http://localhost:8119`:**
`tests/e2e/cashflow-13wk.spec.ts` → **10 passed (2.9m)**, including
`:148 dashboard exposes an Export PDF affordance`.

---

## Local measurements

| check | result |
|---|---|
| PHPUnit full unit suite | base `Tests: 4451, Assertions: 44856, Skipped: 1` → head **`4471 / 44926 / Skipped: 1`**, **0 errors, 0 failures both sides** |
| phpcs `lib/` | exit 0 |
| phpmd `lib/` (both rulesets) | **0 findings** — see below |
| psalm | `No errors found!` |
| phpstan | cannot run locally (`phpstan.neon` points at an uninstalled `nextcloud/ocp` path) — CI is the measurement |
| eslint `src` | 0 errors (96 pre-existing warnings) |
| prettier | clean |
| vitest | 16 files / 177 tests passed |
| `check:manifest` | **PASS with real Ajv**, 228 pages |
| `check:manifest-budget` | PASS — 1,114,863 B of 1,120,000 B |
| `check:registers` / `:seeds` / `:fragment-required` / `:markers` | PASS |
| `test:l10n` | OK |

---

## Two instrument findings worth the reviewer's time

**1. 🔴 `check:manifest` was reporting a pass it had not made.**
`node tests/validate-manifest.js` prints
`Ajv not installed in node_modules … Falling back to a structural lint pass` and
then `structural lint: PASS`. With real Ajv linked in, the **checked-in schema
was stale at 2.11.0** and rejected my action on **seven** counts — its `type`
enum is `handler|open-modal|open-page|navigate` and `additionalProperties` is
`false`, so `api-call` / `url` / `method` / `download` / `filename` /
`successMessage` / `errorMessage` are all refused. nc-vue 2.3.0 — the version
`package-lock.json` resolves — ships schema **2.22.0**, whose `action`
definition declares every one of those fields, and whose runtime
`actionsDispatcher` has implemented them all along. Synced verbatim from
`node_modules/@conduction/nextcloud-vue/src/schemas/`; **all 228 pages pass on
it**. Positive-controlled, because a schema swap that merely widened everything
would also print PASS: with `type: 'total-nonsense-verb'` and a `bogusKey`
planted on the same action, 2.22.0 still reports **both** violations.

**2. `$?` after a pipe is the last stage's code.** My first phpmd run read
`rc=0` through a `| tail` while printing six findings. `phpmd lib text
phpmd.xml` on the whole tree returned **exactly 6**, and **all 6 were mine**
(4× `StaticAccess`, 2× `ShortVariable` `$x`/`$y`) — `development` is
phpmd-clean, so shipping them would have been a regression. `PdfDocument`'s
methods became instance methods injected into the renderer (defaulting to
`new PdfDocument()` so `new CashflowPdfRenderer()` is unchanged) and the
variables were renamed. Now **0**.

---

## Stated, not claimed closed

REQ-CF-016 lists five sections. This ships **§1 (week table)**, **§2 (bar
chart)**, **§3 (assumptions: top-5 customers, recurring breakdown, BTW/IB
methodology)** and **§4 when a scenario is supplied** (unit-covered; production
exports baseline, so no scenario is passed today).

**§5, the "top 3 customers all delay by 14 days" stress test, is NOT built**,
and neither is multi-scenario side-by-side comparison. Both need per-customer,
per-week AR data the renderer's input shape does not carry. I have not marked
them done anywhere.

**A real drift found on the way, adapted rather than renamed:**
`CashflowARProjection` declares `payment_history_average_deviation` while the
renderer reads `gemiddeldeAfwijking`. Nothing had ever called the renderer, so
nothing had noticed. Mapped in the **service** — renaming the renderer's read
key changes a published input contract its own tests pin, and renaming the
schema property is a **data migration**.

---

## Gate parity

Local runner, vendored package `conduction/hydra-gates v1.8.0`, `SCOPE-MODE:
full`. ⚠️ **CI reads the suite from `.github@main`, which moves** — CI's own
numbers are the verdict.

`RESULT: 2 GATE(S) FAILED`:
- **gate-7** `no-admin-idor` — **2 findings, neither is mine**:
  `OssController.php:102 resolveRate` and `SepaAuditController.php:86
  exportMandate`. `CashflowExportController` is **not** in the list. Both files
  are untouched by this diff.
- **gate-19** `e2e-coverage` — 290 scenarios missing `@e2e`, whole-tree debt.
  This PR changes **no** `openspec/**` file.

Base measurement on `origin/development` with the identical package is in the
thread below.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -410,6 +410,15 @@ return \OCA\OpenRegister\AppHost\Routes::standard(
         // to the caller's accessible administrations (IDOR-safe).
             ['name' => 'sepaAudit#exportMandate', 'url' => '/api/sepa/mandates/{mandateId}/audit-export', 'verb' => 'GET'],
 
+        // 13-week cashflow PDF export (zzp-cashflow-13wk, REQ-CF-016 / #865).
+        // POST because the Cashflow Dashboard's declarative `headerActions[]`
+        // api-call action issues a POST through @nextcloud/axios, which carries
+        // the request token — so the endpoint keeps CSRF protection rather than
+        // opting out of it. It takes NO parameters: the horizon is resolved
+        // server-side from the caller's AdministrationMembership set, so no
+        // caller-supplied object identifier crosses the boundary.
+            ['name' => 'cashflowExport#exportPdf', 'url' => '/api/cashflow/export-pdf', 'verb' => 'POST'],
+
         // BCF compensation calculator (bookkeeping-bcf-claim). Pure compute
         // surface returning a what-if compensation result.
             ['name' => 'bcfClaim#compensation', 'url' => '/api/bcf/compensation', 'verb' => 'GET'],
--- a/lib/Controller/CashflowExportController.php
+++ b/lib/Controller/CashflowExportController.php
@@ -1,0 +1,114 @@
+<?php
+
+/**
+ * Cashflow Export Controller
+ *
+ * The route half of REQ-CF-016 (#865): downloads the 13-week cashflow forecast
+ * as a PDF for a bank or accountant meeting. Before this existed,
+ * `CashflowPdfRenderer` had zero callers and `appinfo/routes.php` registered no
+ * cashflow export route at all, so the capability was written and unreachable.
+ *
+ * ## Authorisation posture
+ *
+ * `#[NoAdminRequired]` — this is an operator capability, not an admin one. It
+ * takes NO request parameters: the horizon is resolved server-side from the
+ * caller's own AdministrationMembership set (REQ-MA-001) by
+ * {@see \OCA\Shillinq\Service\CashflowExportService}. There is therefore no
+ * caller-supplied object identifier to guard, which is why no per-object check
+ * appears in this method body — the absence is the design, not an omission.
+ *
+ * No `#[NoCSRFRequired]`: the call is a POST issued by the SPA through
+ * `@nextcloud/axios`, which carries the request token.
+ *
+ * @category Controller
+ * @package  OCA\Shillinq\Controller
+ *
+ * @author    Conduction Development Team <info@conduction.nl>
+ * @copyright 2026 Conduction B.V.
+ * @license   EUPL-1.2 https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * @link https://conduction.nl
+ *
+ * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+ *
+ * SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl>
+ * SPDX-License-Identifier: EUPL-1.2
+ */
+
+declare(strict_types=1);
+
+namespace OCA\Shillinq\Controller;
+
+use OCA\Shillinq\AppInfo\Application;
+use OCA\Shillinq\Service\CashflowExportService;
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\DataDownloadResponse;
+use OCP\AppFramework\Http\JSONResponse;
+use OCP\IRequest;
+use OCP\IUserSession;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * PDF export endpoint for the 13-week cashflow forecast (REQ-CF-016).
+ *
+ * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+ */
+class CashflowExportController extends Controller {
+	/**
+	 * Construct the controller.
+	 *
+	 * @param IRequest $request The request.
+	 * @param CashflowExportService $exportService Assembles the export document.
+	 * @param IUserSession $userSession Session for the auth body-guard.
+	 * @param LoggerInterface $logger Logger.
+	 */
+	public function __construct(
+		IRequest $request,
+		private readonly CashflowExportService $exportService,
+		private readonly IUserSession $userSession,
+		private readonly LoggerInterface $logger,
+	) {
+		parent::__construct(appName: Application::APP_ID, request: $request);
+	}//end __construct()
+
+	/**
+	 * Download the current 13-week cashflow forecast as a PDF (REQ-CF-016).
+	 *
+	 * Answers 404 rather than an empty 200 when the caller's administration has
+	 * no horizon: a blank forecast handed to a bank is a worse artefact than a
+	 * refusal, and the dashboard surfaces the refusal as a toast.
+	 *
+	 * @return DataDownloadResponse|JSONResponse The PDF download, or a JSON error envelope.
+	 *
+	 * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+	 */
+	#[NoAdminRequired]
+	public function exportPdf(): DataDownloadResponse|JSONResponse {
+		if ($this->userSession->getUser() === null) {
+			return new JSONResponse(['error' => 'Authentication required'], Http::STATUS_UNAUTHORIZED);
+		}
+
+		try {
+			$export = $this->exportService->buildHorizonExport();
+		} catch (Throwable $e) {
+			$this->logger->error(
+				'CashflowExportController: cashflow PDF export failed',
+				['exception' => $e->getMessage()]
+			);
+			return new JSONResponse(['error' => 'export_failed'], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		if ($export === null) {
+			return new JSONResponse(['error' => 'no_cashflow_horizon'], Http::STATUS_NOT_FOUND);
+		}
+
+		return new DataDownloadResponse(
+			$export['payload'],
+			$export['filename'],
+			$export['mimeType']
+		);
+	}//end exportPdf()
+}//end class
--- a/lib/Service/CashflowExportService.php
+++ b/lib/Service/CashflowExportService.php
@@ -1,0 +1,337 @@
+<?php
+
+/**
+ * Cashflow Export Service
+ *
+ * Gathers a CashflowForecastHorizon and everything REQ-CF-016 says the export
+ * must contain out of OpenRegister, and hands it to
+ * {@see CashflowPdfRenderer::renderPdf()}. This is the caller the renderer
+ * never had (#865): the renderer existed with ZERO call sites, no route and no
+ * button, so a statutory-reporting capability was written and unreachable.
+ *
+ * ## Scoping, and why the horizon is not addressable
+ *
+ * `buildHorizonExport()` takes NO identifier. It resolves the most recently
+ * rolled horizon belonging to an administration the authenticated caller holds
+ * a valid AdministrationMembership for (REQ-MA-001), through
+ * {@see AdministrationContextService::accessibleAdministrationIds()}.
+ *
+ * That is a deliberate design choice rather than an omission: an endpoint that
+ * accepted a caller-supplied `horizonId` would need its own per-object
+ * authorisation guard, and the dashboard the button lives on
+ * (`/cashflow/dashboard`) carries no object context to supply one from. With
+ * no identifier crossing the boundary there is no IDOR surface to guard. When
+ * a per-horizon export is wanted later it belongs on the horizon DETAIL page,
+ * where the guard has an object to check against.
+ *
+ * ## `filters` addresses JSON properties, and that is the correct use here
+ *
+ * Every lookup below filters on a declared schema property
+ * (`administrationId`, `horizonId`) — never on `id`. `findAll(['filters' =>
+ * ['id' => …]])` matches NOTHING for every value, silently, because the
+ * entity's `id` is its own column and not a property the filter can see; see
+ * {@see \OCA\Shillinq\Util\ObjectIdentifier::findOne()} for the shape to use
+ * when a lookup really is by identifier.
+ *
+ * @category Service
+ * @package  OCA\Shillinq\Service
+ *
+ * @author    Conduction Development Team <info@conduction.nl>
+ * @copyright 2026 Conduction B.V.
+ * @license   EUPL-1.2 https://joinup.ec.europa.eu/collection/eupl/eupl-text-eupl-12
+ *
+ * @link https://conduction.nl
+ *
+ * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+ *
+ * SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl>
+ * SPDX-License-Identifier: EUPL-1.2
+ */
+
+declare(strict_types=1);
+
+namespace OCA\Shillinq\Service;
+
+use OCA\OpenRegister\Contract\ObjectServiceInterface;
+use OCA\Shillinq\AppInfo\Application;
+use OCP\IAppConfig;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * Assembles the REQ-CF-016 cashflow PDF export for the caller's current horizon.
+ *
+ * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+ */
+class CashflowExportService {
+	/**
+	 * CashflowForecastHorizon schema slug.
+	 *
+	 * @var string
+	 */
+	public const SCHEMA_HORIZON = 'CashflowForecastHorizon';
+
+	/**
+	 * CashflowWeek schema slug.
+	 *
+	 * @var string
+	 */
+	public const SCHEMA_WEEK = 'CashflowWeek';
+
+	/**
+	 * CashflowARProjection schema slug.
+	 *
+	 * @var string
+	 */
+	public const SCHEMA_AR_PROJECTION = 'CashflowARProjection';
+
+	/**
+	 * CashflowRecurring schema slug.
+	 *
+	 * @var string
+	 */
+	public const SCHEMA_RECURRING = 'CashflowRecurring';
+
+	/**
+	 * How many customers REQ-CF-016 §3 asks for ("top 5 by AR balance").
+	 *
+	 * @var integer
+	 */
+	private const TOP_CUSTOMERS = 5;
+
+	/**
+	 * Construct the export service.
+	 *
+	 * @param IAppConfig $appConfig App config (OpenRegister register slug).
+	 * @param LoggerInterface $logger Logger — never receives a record body.
+	 * @param ObjectServiceInterface $objectService OpenRegister's object service, injected per ADR-083.
+	 * @param AdministrationContextService $administrationContext Membership guard (REQ-MA-001).
+	 * @param CashflowPdfRenderer $renderer The document renderer.
+	 */
+	public function __construct(
+		private readonly IAppConfig $appConfig,
+		private readonly LoggerInterface $logger,
+		private readonly ObjectServiceInterface $objectService,
+		private readonly AdministrationContextService $administrationContext,
+		private readonly CashflowPdfRenderer $renderer,
+	) {
+
+	}//end __construct()
+
+	/**
+	 * Build the PDF export for the caller's current cashflow horizon (REQ-CF-016).
+	 *
+	 * @return array{filename:string,mimeType:string,payload:string}|null The
+	 *         download envelope, or null when the caller has no accessible
+	 *         administration or that administration has no horizon yet. Null is
+	 *         NOT an empty document on purpose: a blank export of an
+	 *         administration that has never been forecast reads, to a bank, as
+	 *         a business with no cashflow.
+	 *
+	 * @spec openspec/specs/bookkeeping-cashflow-13wk/spec.md#req-cf-016
+	 */
+	public function buildHorizonExport(): ?array {
+		$administrationIds = $this->administrationContext->accessibleAdministrationIds();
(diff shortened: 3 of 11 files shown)


rubenvdlinde:
## Gate parity, both sides measured on the same package

Local runner, vendored `conduction/hydra-gates v1.8.0`, `SCOPE-MODE: full`,
same `vendor/` symlinked into both trees so the environment cannot differ.
⚠️ CI reads the suite from `.github@main`, which moves — **CI's own numbers are
the verdict**; this is the prediction, not the result.

| tree | result | gate-7 | gate-19 |
|---|---|---|---|
| `origin/development` @ `8ea10c02` | `2 GATE(S) FAILED` | FAIL 2 | FAIL 290 |
| this PR @ `6168ee04` | `2 GATE(S) FAILED` | FAIL 2 | FAIL 290 |

**Identical, and gate-7's two are byte-identical findings in files this diff
does not touch:**

```
lib/Controller/OssController.php:102       method=resolveRate      rule=no-auth-guard-in-body
lib/Controller/SepaAuditController.php:86  method=exportMandate    rule=no-auth-guard-in-body
```

`CashflowExportController` is **not** among them — it takes no caller-supplied
identifier, so there is nothing for the gate to find. gate-19's 290 is
whole-tree `@e2e` debt and this PR changes no `openspec/**` file.

Every other gate PASSES on both sides, including gate-1 spdx, gate-3 stub-scan,
gate-5 route-auth, gate-16 spec-coverage, gate-46 spec-anchor-existence,
gate-47 security-change-has-tests, gate-48 csrf-cochange, gate-52
custom-widget-ratchet, gate-53 effective-manifest-crossref, gate-55
detail-page-discipline, gate-57 orphaned-write-capability and gate-60
icon-vocabulary.

---

## The e2e negative control

`tests/e2e/cashflow-13wk.spec.ts:148` is green here. To show it is **my change**
that made it green rather than the rig, I rebuilt the bundle from
`origin/development:src/manifest.json` and deployed only that one file, leaving
the changed `lib/`, the spec file, the fixtures and the instance identical:

| bundle served (md5) | `grep -c export-cashflow-pdf` | `:148 dashboard exposes an Export PDF affordance` |
|---|---|---|
| `743eae82a847e5110afefcf822713f1e` (base manifest) | **0** | **✘ failed (35.7s)** |
| `d7abe16b20c1b2d2c2d6c774a2869cb9` (this PR) | **1** | **✓ passed (17.4s)** |

Both md5s were read back from **inside the container**, not from the build
directory — `js/` is gitignored and the container is a Docker volume, not a bind
mount, so a stale bundle otherwise looks current.

Full spec file on this PR's bundle: **10 passed (2.9m)**.

rubenvdlinde:
## Mutation controls on the export service (head `02987831`)

Three tests added for the branches the first round left uncovered (store
failure, `ObjectEntity`-shaped rows, empty configured register — owed anyway,
and this repo runs a coverage ratchet on the changed-file set).

Every assertion in `CashflowExportServiceTest` is now shown able to fail, **one
change at a time, each reverted before the next**, with `git diff --quiet lib/`
afterwards proving the tree is byte-identical to the commit:

| mutation applied to `CashflowExportService` | result |
|---|---|
| drop the per-administration scoping loop, query unfiltered | `testDoesNotExportAnotherAdministrationsHorizon` **FAILS** |
| reverse the `rolledOn` comparator | `testPicksTheMostRecentlyRolledHorizon` **FAILS** |
| delete the `getObject()` arm of the row normalisation | `testReadsEntityShapedRowsAsWellAsArrays` **FAILS** |
| delete the `weekNumber` `usort` | `testWeeksAreOrderedByWeekNumber` **FAILS** |
| drop the empty-register fallback | `testAnEmptyConfiguredRegisterFallsBackToShillinq` **FAILS** |
| rethrow instead of logging and returning `[]` | `testAStoreFailureIsSwallowedIntoNoHorizon` **ERRORS** |

One caught each; no mutation slipped through.

Two of the new tests do use `createMock(ObjectServiceInterface::class)` — one
configured only to **throw**, one only to **return an entity**. Neither inspects
an argument, so the named-argument blindness cannot produce a false pass. Where
a call's *input* had to be checked (which register slug `setRegister()` was
handed), the test uses a `willReturnCallback` that **records the value and
asserts on the recording**, not a matcher.

28/28 across the three cashflow test classes under both `phpunit-unit.xml` and
CI's `phpunit.xml` (`failOnRisky=true` on both). phpcs and phpmd on `lib/`
still **0**.

rubenvdlinde:
## ⚠️ CI has not returned a verdict yet — the whole shillinq runner queue is stalled

Read this before reading `gh pr checks`.

The run to judge is the **`pull_request` / Code Quality** run at the current
head `02987831`: **`32021200293`**. As of this comment it is
`status: pending` with **`total_count: 0` jobs** — it has not started, which
is **no verdict**, neither pass nor fail.

This is not specific to this PR. The repo's whole queue is backed up: a
`development` **Code Quality** run created at `09:38:42Z` is *also* still
pending, as are `CodeQL`, `Development Release` and `OpenSpec Sync`.

Two things `gh pr checks` will show that are **not** results:

- run `32020127509` (**push**, Code Quality) renders ~20 jobs as `fail` at
  `2m12s` / `0s` — it was **cancelled by the concurrency group** when the PR
  opened. A cancelled twin is not a failure.
- run `32020299728` (**pull_request** @ `6168ee04`) is `cancelled` too, by the
  follow-up push.

Verdict command:

```
gh api repos/ConductionNL/shillinq/actions/runs/32021200293/jobs?per_page=100
```

Assert `total_count > 0` before concluding anything from it.

Local and on-instance measurements are in the PR body and the two comments
above; the gate-parity prediction (base and head both `2 GATE(S) FAILED`, same
two gate-7 findings in files this diff does not touch) is the one CI should be
checked against.

rubenvdlinde:
## 🔴 CI verdict: 3 gates fail that `development` passes — one is mine and fixed, two are a gate-package defect fixed at source

**Read this before `gh pr checks`.** Run `32021200293` (`pull_request`, Code
Quality @ `02987831`): **33 success / 3 failure / 4 skipped**.

The base is the comparison that matters: **`development` @ `8ea10c02` has
`Hydra Gates: SUCCESS`** (job `95374987863`). So this was a **real regression**,
not parity. It is not one any more; here is each one.

⚠️ **My local gate run predicted the wrong gates entirely.** I measured
`gate-7 FAIL 2` / `gate-19 FAIL 290` on the vendored `conduction/hydra-gates
v1.8.0`. **CI ran package `ac5aec11`**, on which gate-7 **PASSES** and gate-19
is **WARNING (advisory, non-blocking)** — and three *different* gates fail. The
package moving under you is a documented trap and it caught me: a local gate
number is worthless unless it names the same package sha CI used.

### gate-60 icon-vocabulary — MINE, FIXED at `2321c902`

```
FAIL  src/icons.js: 1 icon name(s) used by the manifests are NOT registered —
      they render with no icon at all, not a fallback: FilePdfBox (ADR-077 rule 3).
```

`FilePdfBox` exists in `vue-material-design-icons` but shillinq's `icons.js`
never imported it, so the button would have rendered with **no icon**. The gate
is right. Switched to **`DatabaseExportOutline`** — already registered here, and
the canonical concept icon the gate's own `semantic-icons.json` maps `export`
to, so it cannot raise a Tier-B SHOULD warning either. Measured with the gate's
own helper on both trees: **1 failure → 0**, warnings **unchanged at 22**, so no
new warning was traded in for the fix.

### gate-22 + gate-53 — ONE root cause, and it is not shillinq's

Both report the **same** error at the **same** pointer — gate-53 validates the
*assembled* manifest, so it inherits gate-22's finding:

```
at /pages/142/config/headerActions/0: must NOT have additional properties
```

The gate package vendors `app-manifest-v2.schema.json` at **2.19.0**, whose
`action` definition is `additionalProperties: false` and declares no
`download` / `filename`. But **nc-vue 2.3.0's `actionsDispatcher` implements
exactly those** (`responseType: 'blob'`, filename off `Content-Disposition`),
and **nc-vue's own schema 2.22.0 declares them**. The vendored copy's `$id`
points at nc-vue's `main`, so it is a mirror that went three minor versions
stale.

**There is no in-vocabulary alternative, checked rather than assumed:**
`navigate` is a vue-router `push` (no document navigation ⇒ no download) and
`export` opens `CnMassExportDialog`, not a file save. Dropping `download: true`
leaves an action that POSTs and toasts and **never downloads** — i.e. exactly
the "looks wired, isn't" shape #865 was filed about.

**Fixed at source: ConductionNL/.github#494**, syncing the vendored schema to
2.22.0. Measured strictly more permissive (nothing lost: no def, no action
property, no enum value), both directions (shillinq FAIL 1 → PASS; procest /
docudesk / openbuild / opencatalogi / softwarecatalog / decidesk PASS → PASS),
and positive-controlled — a planted bad action is still rejected on both counts.

**This PR is therefore blocked on #494**, not on anything in shillinq. Once
`.github@main` carries it, gates 22 and 53 clear with no further change here.
I have deliberately **not** contorted the manifest to satisfy a stale schema.

---

## E2E: 8 → 7, and the only spec that changed state is #865's

By title, from both job logs (byte-counted: 343,550 and 339,345):

| spec | base `8ea10c02` | PR |
|---|---|---|
| `cashflow-13wk.spec.ts:148` Export PDF affordance | ✘ | **✓ GONE** |
| `provincies-bbv-variant.spec.ts:145 / :165 / :194 / :212 / :313` | ✘ ×5 | ✘ ×5 |
| `spec-coverage/inventory.spec.ts:61` Products / Product Attributes | ✘ ×2 | ✘ ×2 |

**8 → 7. No new failure, no swapped flake, no title I did not expect.** The
remaining seven are #866/#862 and #860 — the untouched product gaps.

## Everything else on CI

**33 jobs green**, including all six PHPUnit matrix cells, Newman, `Frontend
Build`, `Coverage Baseline Protection`, `check:manifest`, `check:manifest-budget`
and every static analyser — **`phpstan` among them**, which cannot run locally
here. The 4 skips (`Features Extract`, `Nextcloud API check`, `SBOM`,
`Journeydoc Capture`) are the same four that skip on the base; a skip is no
verdict.

rubenvdlinde:
## ✅ CI SETTLED — at parity with `development`, and strictly better

`pull_request` / Code Quality run **`32026875818`** @ **`2321c902`**:
**35 success / 2 failure / 4 skipped.**

| | base `development` @ `8ea10c02` (run `32016271020`, completed) | PR #901 @ `2321c902` |
|---|---|---|
| `E2E Tests (Playwright)` | **failure** — 8 failed / 6 skipped / 259 passed | **failure** — **7 failed / 6 skipped / 260 passed** |
| `Quality Report` | **failure** (pure aggregator over E2E) | **failure** (same) |
| `Hydra Gates` | success | **success** |
| everything else | success | success |

**The PR's failing job set is identical to the base's ⇒ at parity under the
merge policy, and strictly better inside E2E: one fewer failure, one more pass,
the same six skips.**

## Hydra Gates is fully green

```
[gate-22] manifest-validation: PASS
[gate-53] effective-manifest-crossref: PASS   (5 WARN, non-blocking)
[gate-60] icon-vocabulary: PASS
[hydra-gates] RESULT: ALL 64 APPLICABLE GATES PASSED — and all 64 of them ran.
```

`SCOPE-MODE: full`, so this is not a zero-scope pass.

**Attribution, because two different things cleared three gates and only one of
them was in this PR:**

- **gate-60** cleared by this PR's `2321c902` (`FilePdfBox` →
  `DatabaseExportOutline`, the registered canonical `export` icon).
- **gate-22 + gate-53** cleared because **ConductionNL/.github#494 merged** at
  `12:03:55Z` as **`97bb374b`** — and `97bb374b` is *exactly* the gate package
  sha this CI run reports. Verified rather than inferred:
  `git show 97bb374b:hydra-gates/scripts/schemas/app-manifest-v2.schema.json`
  reads `version: 2.22.0`, `download` present. **Nothing in shillinq changed to
  make those two pass** — the stale vendored schema did. I opened #494; I did
  not merge it.

⚠️ **The gate package moved twice during this one PR** — `ac5aec11` →
`97bb374b` — on top of my local vendored `v1.8.0` disagreeing with both. Any
gate number that does not name its package sha is not a measurement.

## E2E, by title, at this head

7 failed / 6 skipped / **260 passed** (base: 8 / 6 / 259). Log byte-counted at
340,790 before reading.

```
provincies-bbv-variant.spec.ts:145 / :165 / :194 / :212 / :313   (#866/#862)
spec-coverage/inventory.spec.ts:61  Products / Product Attributes (#860)
```

`cashflow-13wk.spec.ts:148 dashboard exposes an Export PDF affordance` is
**gone from the failure list** — the one spec that changed state, in the right
direction, by being satisfied. No new failure and no swapped flake; the
intermittent-9th-failure risk did not materialise on either side.

**Ready for review. Not merged — the verifier merges.**

PR #943, opened 2026-08-20
build(deps-dev): bump stylelint to 17 as one coupled upgrade (supersedes #898, #897, #891)

rubenvdlinde:
## Why one PR instead of several

`stylelint` and its config packages cannot move independently — the config packages declare `stylelint >=16` as a peer. Each single-package dependabot PR left the tree unresolvable, and stylelint aborted **before linting a single file**:

```
Error: Could not find "stylelint-config-recommended-scss".
##[error]Process completed with exit code 78
```

Those PRs never failed on CSS. They failed on an unresolvable dependency tree.

## What changed

| package | to |
|---|---|
| `stylelint` | ^17.14.1 |
| `@nextcloud/stylelint-config` | ^3.2.2 |
| `stylelint-config-recommended-scss` | ^17.0.1 |
| `stylelint-config-recommended-vue` | ^2.0.0 |
| `stylelint-config-html` | ^2.0.0 *(newly declared)* |

`stylelint-config-html` is a peer of `stylelint-config-recommended-vue@2`. The old tree happened to hoist it; the new one does not, so it is declared explicitly rather than relied on transitively.

## CSS fixes

stylelint 17 promotes two deprecated-CSS rules to **errors**. Both are genuine deprecations, not rule noise, and are fixed here:

- `word-break: break-word` → `overflow-wrap: break-word` — `break-word` was a legacy alias that always behaved as `overflow-wrap`; the rendering is unchanged.
- `clip: rect(0, 0, 0, 0)` → `clip-path: inset(50%)` — in the visually-hidden / screen-reader-only blocks. `clip` is deprecated; `clip-path: inset(50%)` is the standard replacement and clips identically.

## Verification

`npm run stylelint` exits **0** locally on node 22, lockfileVersion 3 (unchanged, so CI's npm can read it).

Remaining `csstools/use-logical` findings are **warnings** — they do not fail the gate and are deliberately left for a follow-up so this stays a scoped dependency upgrade.


rubenvdlinde:
Merged current `development` in (the branch was **103 commits behind, 1 ahead**). Clean merge, no conflicts.

The E2E failures on the previous run were not caused by this change. A stylelint bump cannot break page rendering, and this branch predated the SPA boot fix — with the entry point not loading its own split chunks, essentially every page-level spec failed regardless of branch contents. Re-running against current `development` is what makes this PR's own verdict meaningful.

Contents are otherwise unchanged: the stylelint 17 family (`stylelint`, `stylelint-config-recommended-scss`, `stylelint-config-recommended-vue`, `stylelint-config-html`, `postcss-html`) moving together, plus the three deprecations stylelint 17 flags — `word-break: break-word` → `overflow-wrap: break-word` (×2) and `clip: rect(0,0,0,0)` → `clip-path: inset(50%)`.

Also closed #891, #897 and #898, the bare single-package bumps this supersedes: each of those configs peers on `stylelint`, so they cannot land one at a time.

rubenvdlinde:
Merged current `development` in again — the branch was **35 commits behind** and its E2E failure was not its own.

It was still failing `cbs-submissions.spec.ts:220`, which `development` has since fixed (the confirm button is in NcDialog's `#actions` slot, a sibling of the `data-testid-phase` marker, so the old locator matched zero buttons). This branch simply did not have that fix yet — 0 occurrences of `button-vue--error` on the old head, 2 now.

Verified with the commands CI runs, not proxies for them:

```
npm ci                                        rc=0
npm run stylelint    ⚠ 59 problems (0 errors) rc=0
prettier --check     All matched files use Prettier code style!
```

`npm ci` matters here specifically: it refuses peer conflicts that `npm install` only warns about, and this PR moves the whole stylelint peer family together.

PR #955, opened 2026-08-20
fix(e2e): isolate the spec that resets server state (5 failures in a 33s window)

### LINKED ISSUE #928: e2e: 5 Playwright tests fail on development since 2026-08-19 (Budget-to-Programme Linker, KOR, Tax Filing Prep, setup wizard)
## 5 Playwright e2e tests fail on `development` (261 pass)

E2E on `development` has failed on every run since **2026-08-19T22:16**. Last green: **2026-08-18T21:58**. The failing set was 7, is now 5 — so it is partly self-healing, not static.

### Failing tests

| Test | Assertion |
|---|---|
| `provincies-bbv-variant.spec.ts:274` | `getByTestId('cn-index-page')` never appears |
| `provincies-bbv-variant.spec.ts:313` | same |
| `spec-coverage/belastingen.spec.ts:106` — KOR (`/belastingen/kor`) | renders **no** index surface: `tables=0 empty=0 lists=0 actions=0` |
| `spec-coverage/belastingen.spec.ts:106` — Tax Filing Prep (`/tax-filing-prep`) | same |
| `setup-wizard-english.spec.ts:240` | wizard/shell gating |

### What I ruled out

- **Not the dependency bumps.** Six majors landed together at 21:23–21:24 (`@nextcloud/files` 4.0.0, `gridstack` 13, `plugin-vue` 6.0.8, `webpack-vue-config` 7.0.4, `postcss-html` 2, `compiler-sfc` 3.5.41). Other fleet apps took comparable bumps and their e2e is green — the breakage is shillinq-only.
- **Not a missing route.** `/belastingen/kor` (`KorStatus`) and `/tax-filing-prep` (`TaxFilingPrep`) are both present in `src/manifest.json` as well-formed `type:"index"` pages with `config.register`, `config.schema` and `columns`.
- **Not the lazy-fragment loader.** `/tax-filing-prep` ships in the bundled `manifest.json` with full config, so it is never lazy-loaded via `require.context('./manifest.d/', … 'lazy')` — yet it still fails.
- **Not the slot component.** `BbvLinkerFilterBar` exists and is registered in `src/registry.js`; its endpoint `/api/bbv-provincie/gl-line-facets` is routed (`appinfo/routes.php:436`) and the controller method carries `#[NoAdminRequired]` plus a login check.
- **Not a changed visibility gate.** `KorStatus.config.visibility` (`administrationType in [mkb, zzp]`) is byte-identical at the last-green commit `dfbc7b9`, and `TaxFilingPrep` has no visibility gate at all.

### Leads worth pulling

1. **Duplicate manifest entries.** `src/manifest.json` contains **two** `KorStatus` page objects — one with `config.visibility` and one without. Last-one-wins would hide the page. This predates the regression but may interact with something that changed.
2. **Two disjoint manifests.** `src/manifest.json` declares 228 routes, the `src/manifest.d/*.json` fragments declare 375, and **218 built routes are absent from the fragments**. `main.js` merges `bundledManifest` + `manifestShell.fragments`. `/bbv-provincie/budget-to-programme` exists **only** in the fragments; `/belastingen/kor` and `/tax-filing-prep` **only** in the bundled manifest. Both halves have failures, so the merge itself is the common suspect.
3. **`type:"report"` legacy.** `TaxFilingPrep`'s own description says it was "converted from the never-rendered `type=report` (no such page type is registered)".

### Next step

The traces are the fastest route — each failure uploaded one:
`npx playwright show-trace tests/e2e/test-results/<test>/trace.zip` from the `playwright-report` artifact of the failing run.

_Filed automatically during a fleet merge session; analysis is static only — no runtime reproduction was performed._

rubenvdlinde:
Closes #928.

## The five failures have two causes, and neither is the page they point at

All five land inside a **33-second window** (run `32329757674`):

| time | spec |
|---|---|
| 04:05:50 | `provincies-bbv-variant.spec.ts:274` |
| 04:06:09 | `spec-coverage/belastingen.spec.ts:106` (KOR) |
| 04:06:16 | `provincies-bbv-variant.spec.ts:313` |
| 04:06:18 | `spec-coverage/belastingen.spec.ts:106` (Tax Filing Prep) |
| 04:06:23 | **`setup-wizard-english.spec.ts:240`** ← the resetting spec |

`setup-wizard-english.spec.ts` deletes shillinq's setup app-config keys (`legal_country`, `administration_id`, `setup_completed_version`, …) with `occ config:app:delete` to prove the wizard renders on a genuinely first-run instance, then restores the `ci-seed.sh` baseline in `afterAll`. Files run in **parallel** (`workers: 4`).

For the length of that reset, every sibling spec gets a blocking "Set up this app" modal over `main`. That is exactly what the four victims' page snapshots show — the wizard open at step 5 with its Run button active, and nothing behind it:

```yaml
- dialog:
  - heading "Set up this app"
  - tab "5 Create administration" [selected]
  - button "Run" [active]
```

So `tables=0 empty=0 lists=0 actions=0` was accurate and precise: **not an empty index — no index.** The four "failing pages" are fine.

The spec's own header names this hazard ("running this file with `--workers=1` … is still the safer choice"). Nothing enforced it.

## Fix 1 — isolate the resetting spec

`setup-wizard-english.spec.ts` joins `CI_TEST_IGNORE` and gets an **opt-in** project gated on `SHILLINQ_SETUP_WIZARD_SPEC=1`:

```
SHILLINQ_SETUP_WIZARD_SPEC=1 npx playwright test \
  -c tests/e2e/playwright.config.ts --project=setup-wizard --workers=1
```

The env gate is load-bearing: a declared project is part of a default (no `--project`) run, so declaring it unconditionally would put the spec straight back beside its victims and undo the isolation.

> **Do not "fix" this by dismissing the dialog in `dismissOverlays()`.** My first commit on this branch did exactly that and it is reverted here — a victim completing the wizard re-runs `init-administration` underneath this spec's own walk and corrupts the state it is asserting. `_helpers.ts` now carries a comment saying so.

## Fix 2 — a genuine strict-mode collision

`setup-wizard-english.spec.ts` also fails on its own merits:

```
strict mode violation: getByRole('dialog').first().getByText('Welcome to Shillinq')
resolved to 2 elements:
  1) <span class="cn-wizard-dialog__progress-label">Welcome to Shillinq</span>
  2) <p class="notecard__heading">Welcome to Shillinq</p>
```

The wizard grew a progress tablist whose labels are the step titles, so every title now appears twice. All **seven** step assertions collide — the run only reported the first because the assertion throws. They now target `.notecard__heading`, the step body this spec is about.

## Verification

`--list`, with controls in both directions:

| | result |
|---|---|
| default run collects `setup-wizard-english` | **0** |
| default run still collects victim tests | **30** |
| `--project=setup-wizard` collects it | **1** |
| …with other specs leaked in | **0** |

`tsc --noEmit --skipLibCheck` clean on all three files. The behavioural proof is CI on this PR.

## Follow-up worth doing separately

CI runs only the default project, so `setup-wizard-english.spec.ts` is now **not executed** until a job invokes the opt-in project. That is strictly better than it silently breaking four other specs, but it is a coverage gap someone should close with a second step in the workflow.


rubenvdlinde:
**Confirmed by measurement, and the prediction held.**

| state | e2e failures |
|---|---|
| `development` before #968 | **~190** — SPA did not boot at all |
| after #968 (load the shared chunks) | **14** |
| **this PR on top** | **3** |

The 3 that remain are exactly the ones named in advance when #968 merged — "after both land, the expected residual is the 3 unrelated pre-existing failures":

```
cbs-submissions.spec.ts:149   CBS Submissions — list view
cbs-submissions.spec.ts:195   CBS Submissions — delete-own-draft flow
external-adapters.spec.ts:207 External Connections roster
```

Neither `setup-wizard-english` nor any `spec-coverage` victim appears any more. That is the contamination cluster gone: isolating the one spec that deletes server-side setup config removed ~11 failures across `belastingen`, `bookkeeping` and `cashflow-pension`, plus the wizard's own strict-mode collision.

Merging on **parity, not green** — 3 failures against `development`'s 190, and all 3 predate both changes and have a different cause.

## Known coverage gap, deliberately left visible

CI runs only the default project, so `setup-wizard-english.spec.ts` is now **not executed** until a job invokes `--project=setup-wizard --workers=1`. That is strictly better than it silently breaking eleven other specs, but it is a gap someone should close with a second workflow step.

rubenvdlinde:
Closing — `development` already carries an equivalent fix, arrived at independently.

`fddd3c350 fix(e2e): isolate the setup-wizard spec — its server-side reset was failing 4 other spec files` implements the same two things this PR does:

- an **opt-in** project for `setup-wizard-english.spec.ts`, gated on `RUN_SETUP_WIZARD_SPEC` (this PR used `SHILLINQ_SETUP_WIZARD_SPEC`), excluded from the default run so it cannot sit beside the specs it resets;
- the same **strict-mode** scoping fix for the step titles, which now collide with `CnWizardDialog`'s progress tablist.

Same diagnosis, same shape, different variable name. There is no reason to merge a second implementation, and the conflict between them is purely the two spellings.

**What this PR measured is still worth recording**, since it was taken against a tree that had only one of the two fixes at a time:

| state | e2e failures |
|---|---|
| `development` before #968 | ~190 — SPA did not boot |
| after #968 (load the shared chunks) | 14 |
| with this isolation on top | **3** |

The 3 that survive are `cbs-submissions.spec.ts:149`, `cbs-submissions.spec.ts:195` and `external-adapters.spec.ts:207` — unrelated to both changes and pre-dating them. That is the number to expect on `development` now that #968 and `fddd3c350` are both in.

Superseded by `fddd3c350`; #968 carried the part that was not duplicated.

PR #968, opened 2026-08-20
fix(build): load the shared chunks #932 split out — the SPA does not boot on development

rubenvdlinde:
**The shillinq SPA does not boot on `development`.** ~190 e2e tests — nearly the whole suite — fail with one shape:

```
Error: expect(locator).toBeVisible() failed
Locator: locator('main, [role="main"]')
Error: element(s) not found
```

## Cause

#932 (`frontend-bundle-hygiene`, ADR-061) split Vue, `@nextcloud/vue`, `@conduction/nextcloud-vue` and Pinia out of the entries into shared chunks with **`chunks: 'initial'`** — but left the templates naming one script each.

An **initial** chunk is not fetched by the webpack runtime. The page must request it, and this app has no `HtmlWebpackPlugin` to inject the tags. Webpack's own build output states the entry points are three files:

```
Entrypoint main          = shillinq-shared-nc-vue.js  shillinq-shared-vendor.js  shillinq-main.js
Entrypoint adminSettings = shillinq-shared-nc-vue.js  shillinq-shared-vendor.js  shillinq-settings.js
```

Both chunks **are** emitted — 973 KiB and 51.1 KiB (run `32358287419`). Only one of the three was ever loaded, so the framework never arrived:

| | before #932 | after |
|---|---|---|
| `shillinq-main.js` | 12,468,786 B | **9,158,113 B** |
| e2e failures | 5 | **~190** |

**Frontend Build stayed green the whole time.** Emitting a chunk nobody requests is not a build error — only e2e could see this, and #932 was merged with its E2E job already red.

## The fix

The **missing half of #932**, not a revert. The optimisation is kept; the templates now load the three files in dependency order:

```php
Util::addScript($appId, $appId . '-shared-vendor');
Util::addScript($appId, $appId . '-shared-nc-vue');
Util::addScript($appId, $appId . '-main');       // and '-settings' in the admin template
```

This is exactly what `pipelinq/templates/index.php` already does — and #932 cited that file as its model, but copied only the webpack half.

`widget` needs nothing: `webpack.config.js` excludes it from splitting (`chunks: (chunk) => chunk.name !== 'widget'`) so the UMD embed bundle stays self-contained for third-party pages.

## Verification

- Both shared chunks confirmed **emitted** in the CI build output, with webpack naming all three files per entrypoint.
- `php -l` clean on both templates.
- The real proof is e2e on this PR: if the shared chunks now load, `main` mounts and the ~190 failures collapse back to the 5 pre-existing ones (a separate cause, addressed in #955).

## Note on the first commit

This branch opens with a revert of the splitChunks block. That was my first fix, before I checked whether the chunks were emitted and how pipelinq handles it. The second commit reverts that and does it properly — the history is left intact rather than rewritten.

## Fleet check

hermiq and portaliq also declare `shared-nc-vue` / `shared-vendor` chunks and already carry 4 `addScript` lines, so they are not affected. openregister, openbuild, decidesk and scholiq have no initial-chunk split at all.

### DIFF
--- a/templates/index.php
+++ b/templates/index.php
@@ -3,6 +3,27 @@
 use OCP\Util;
 
 $appId = OCA\Shillinq\AppInfo\Application::APP_ID;
+// The webpack build (see webpack.config.js -> optimization.splitChunks, added
+// by #932) emits this entry point as THREE files, not one. Webpack's own build
+// output says so:
+//
+//     Entrypoint main = shillinq-shared-nc-vue.js
+//                       shillinq-shared-vendor.js
+//                       shillinq-main.js
+//
+// All three must be loaded, in dependency order — `shillinq-main.js` references
+// modules that now live in the shared chunks. An `initial` chunk is NOT fetched
+// by the webpack runtime; the page has to request it, and this app has no
+// HtmlWebpackPlugin to inject the tags.
+//
+// Loading only `-main` is what broke the SPA on `development`: the entry went
+// 12,468,786 -> 9,158,113 bytes, the framework never arrived, and ~190 e2e
+// tests failed with `locator('main, [role="main"]')` never appearing. Frontend
+// Build stayed green throughout — emitting a chunk nobody requests is not a
+// build error. pipelinq/templates/index.php carries the same three lines for
+// the same reason.
+Util::addScript($appId, $appId . '-shared-vendor');
+Util::addScript($appId, $appId . '-shared-nc-vue');
 Util::addScript($appId, $appId . '-main');
 ?>
 <div id="shillinq-app"></div>
--- a/templates/settings/admin.php
+++ b/templates/settings/admin.php
@@ -3,6 +3,17 @@
 use OCP\Util;
 
 $appId = OCA\Shillinq\AppInfo\Application::APP_ID;
+// Same three-file entry point as templates/index.php — webpack's build output:
+//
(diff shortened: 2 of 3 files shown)


rubenvdlinde:
**Merging on PARITY, not green — with the numbers, because `development` is currently far worse.**

e2e on this branch (run `32362969761`): **14 failures**.
e2e on `development` right now: **~190 failures** — nearly the whole suite, all `locator('main, [role="main"]')` never appearing, because the SPA does not boot at all.

So this is not "red PR merged anyway": it restores a booting application and takes the suite from ~190 failures to 14.

## What the remaining 14 are — none of them this change

| count | spec | cause |
|---|---|---|
| ~10 | `spec-coverage/{belastingen,bookkeeping,cashflow-pension}` | victims of `setup-wizard-english.spec.ts` deleting server-side setup config while they run in a parallel worker — the 33-second-window contamination diagnosed in #955 |
| 1 | `setup-wizard-english.spec.ts:240` | Playwright strict-mode collision, also fixed in #955 |
| 3 | `cbs-submissions` (2), `external-adapters` (1) | separate, pre-existing |

This branch is cut from `development`, so it does **not** carry #955's isolation fix — which is exactly why the contamination cluster is still present here. #955 removes it; the two changes are complementary and neither is blocked by the other.

After both land, the expected residual is the 3 unrelated pre-existing failures.

PR #998, opened 2026-08-20
fix(manifest): process and audit fields are no longer editable inputs

rubenvdlinde:
Part of the fleet data-exposure sweep (see ConductionNL/openregister#2644 for the half that cannot be fixed in a manifest).

## The problem

`CnObjectDataWidget.editable` defaults to **`true`**, so every property named in a data widget's `include` list becomes a text box. That put lifecycle state and audit stamps in front of users as editable fields: `status`, `lifecycle`, `submittedAt`, `approvedBy`, `openedAt`, `closedAt`, `publishedAt`, `enactedAt`.

Those are written by the backend when a transition lands — `TransitionEngine` stamps them through `saveObject()` — so an input for them is a control that can only fail or confuse. The guarded path is the lifecycle buttons; `LifecycleValidationListener` rejects anything that is not a legal transition.

## The fix

Per-field `overrides.<field>.editable: false`, **not** `editable: false` on the widget. These panels mix process state with fields the user legitimately edits, and a blanket lock would make those read-only too.

## What this does not fix

Widgets that declare **no** `include` render *every* schema property. Enumerating their fields in a manifest would drift the moment the schema changes — 52 such widgets fleet-wide expose 124 process fields. Closing those needs a server-side "system-owned" marker that OpenRegister does not have: `readOnly: true` has no bypass for backend callers and would reject the very transition that writes the field. Filed as ConductionNL/openregister#2644.

## Verification

Inserted textually — one compact line per widget — so the diff is the change and nothing else. A full JSON re-serialisation reflowed hand-compacted lines and turned this into thousands of lines of churn on first attempt; this version is a handful of lines per app.

A verifier re-parses both files and asserts the only structural difference is the added overrides, that no non-process field was locked, and that no widget definition moved. `check:manifest` passes.


rubenvdlinde:
## E2E triage — not this change

**300 passed, 4 failed:**

- `budget-core-schema.spec.ts` — Budgets nav group, LedgerGroup detail children, BudgetLine 12 monthly amounts
- `cbs-submissions.spec.ts` — delete-own-draft flow

This PR changes **one field**: `overrides: { "status": { "editable": false } }` on the `profile-schedule` widget of `RecurringInvoiceProfileDetail`, in `src/manifest.d/recurring-invoicing.json`.

No overlap with either spec. `cbs-submissions.spec.ts` never mentions recurring invoicing; `budget-core-schema.spec.ts`'s single match is a **comment on line 41** citing `invoice-quick-draft.spec.ts` as a precedent, not a dependency.

The run also carries a schema/migration mismatch in the test environment, repeated throughout:

```
ERROR: column "emu_inclusion_rule" does not exist at character 167
ERROR: relation "openregister_vec_ann" does not exist at character 59
```

I could not establish parity from `development` — **every one of its last 20 runs is concurrency-cancelled**, which is the release-loop cadence outrunning CI rather than anything about this branch. So this is a causal argument rather than a parity one: the change cannot reach the failing subsystems.

Worth a second opinion if you disagree — but I would not hold this PR for it.

PR #1002, opened 2026-08-20
chore(quality): migrate to PHPStan 2 and clear all 115 residual errors

rubenvdlinde:
## Result

`phpstan analyse`: **115 errors → 0**. **4748 PHPUnit tests green** (45,624 assertions, 1 pre-existing skip). `phpcs`: **0 errors**.

Bumps `phpstan/phpstan` to `^2.0` and `conduction/hydra-gates` to `^1.8.2`. This was the largest of the nine app migrations; the findings fell into a handful of repeated shapes plus a few that were genuinely worth reading closely.

## Dead guards around the OpenRegister contract (93 of the 115)

`ObjectServiceInterface::saveObject()` is declared `: ObjectEntityInterface` and `findAll()` `: array`; `ObjectEntityInterface` extends `JsonSerializable` and declares `getObject(): array`.

**I verified that against the canonical `openregister` contract on both `main` and `development`, not against this app's own stub.** That distinction matters here: `tests/stubs/OpenRegister/Contract/` is hand-maintained in this repo, so checking a guard against it would only confirm what we wrote ourselves. (The stub and the canonical interface do agree — both declare the same six methods and extend `JsonSerializable`.)

Given that:

- **12** `is_object()` / `method_exists()` pairs guarding `jsonSerialize()`/`getObject()` across six near-identical `saveOnSchema()` helpers, each with an unreachable fallback arm and an unreachable trailing `throw`.
  **The `is_array()` check on `jsonSerialize()` stays** — that method returns `mixed`, so it is the one check in the block still doing work.
- **51** `is_array()` guards on `findAll()` results — as early returns, as `|| $x === []` conjuncts, and as blocks that only ever assigned the value straight through.
- `method_exists($activity, 'setMessage')` on an `OCP\Activity\IEvent`.

Two `use RuntimeException;` imports lost their last reference along with the throws, and are removed.

## Redundant conditions (22)

Nineteen are `isset(...) && ... !== null` — `isset()` is already false for null. Three needed their own reasoning:

- **`BadoControleprotocolService`** — `$sampleIds !== []` sits after an `empty($sampleIds)` early return.
- **`SettingsService`** — `$dedupeKey !== ''` where the narrowed type is `'code'|'reference'`.
- **`WbsoAccountService`** — `while ($cursor !== '')`. The loop only ever reassigns `$cursor` from a `$next` it has already refused when empty, so the condition could never end the loop. It is now `while (true)`, which is what the control flow has always been; the `break` and the throws are the real exits.

## Numeric-string array keys — two docblocks describing something PHP does not do

`KapitaallastenCalculator::schedule()` documented `array<string,float>` and cast each year with `(string)`. `BankRulePreviewService::previewRule()` documented `predicateBreakdown: array<string,int>` and cast each predicate index the same way.

**PHP coerces a numeric string array key back to an integer.** Neither cast could ever have produced string keys, so both docblocks were wrong and both casts were no-ops. Docblocks corrected to `array<int,...>`, casts removed.

Callers indexing with `'2027'` keep working — the *lookup* is coerced the same way. That symmetry is exactly why this went unnoticed, including by the tests, which index with string literals and pass either way.

## Six unresolvable JSONResponse payloads

- **`CBSSubmissionController::objectService()`** returned a bare `object` while the injected property was already `ObjectServiceInterface` — the accessor was throwing the type away. Now returns the interface.

- **`DBAController::objectService()`** resolves the service by container id. Its native return type **stays `?object` on purpose**, and the narrowing lives in the docblock instead.

  Worth being explicit, since it looks like the weaker choice: I first added a runtime `instanceof ObjectServiceInterface` here. It turned **22 passing tests into 503s**, because the unit tests inject a focused fake implementing the four methods this controller actually calls rather than all 26 on the contract. That fake is reasonable, and rejecting it would not have made production any safer — the real container returns the real service either way. So the docblock buys the type resolution without dictating the runtime.

- The sixth was `($validation['valid'] ?? false) === false` on a value declared `array{valid:bool,errors:...,warnings:...}`. The `?? false` on an always-present key made PHPStan lose the shape *inside that branch*, leaving `$validation['errors']` unresolved and the entire response payload unverifiable. Dropping it restores the shape.

## Remaining mechanical fixes (4)

Three `array_values()` calls on values that are already lists (two of them wrapping `array_keys()`), and one `?? null` on offset 0 of a non-empty list.

## Review note

The six collapsed `saveOnSchema()` helpers are the biggest behavioural surface here. Each previously ended in `throw new RuntimeException('… unsupported row type …')`, which is now unreachable and removed. If you would rather keep those throws as belt-and-braces against a future contract change, say so and I will restore them behind an explicit `@phpstan-ignore` with the reason recorded — but as the contract stands on both branches, they cannot fire.


rubenvdlinde:
Merged current `development` in — the branch was **25 commits behind**, and the E2E failures it hit were not its own.

Three specs failed: `budget-core-schema.spec.ts:117`, `:298` and `cbs-submissions.spec.ts:220`. The same three fail on **#967**, an entirely unrelated branch (the budget feature). Two independent branches failing the identical set points at `development`, not at either change — and this PR touches only PHP under `lib/`, none of it the budget nav or the CBS delete flow.

The timeline confirms it:

- `budget-core-schema.spec.ts` landed on `development` at **2026-08-20 09:20**.
- `fix(e2e): BudgetLine skip guard checked visibility instead of options` (#996) landed at **22:44**.
- This PR's failing run started at **22:47**, from a merge base that predates #996.

So the branch was being judged against a known-broken snapshot that had already been fixed upstream. Re-running against current `development` is what makes this PR's verdict its own.

After the merge, locally: **phpstan 0 errors**, **4784 PHPUnit tests green** (45,701 assertions, 1 skipped).

Also added in the previous commit: `@spec` anchors for the three methods gate-16 flagged — `BillingModelEngine::calculateTAndM` to the *Billing Model Logic (T&M, …)* requirement, and the two `WbsoAccountService` hierarchy methods to chart-of-accounts **REQ-CoA-003**. I did not reuse the class-level anchor those files carry: it points at `openspec/changes/bookkeeping-wbso-sno-administratie/`, a change directory that no longer exists on disk.

rubenvdlinde:
Merging on **parity**, not green — with the evidence for that, because the base is red.

### The 3 E2E failures are not this PR's

```
budget-core-schema.spec.ts:224   P&L-shaped LedgerGroup rows are present, and no balance-sheet row is
budget-scenarios.spec.ts:398     create a LEDGER_AMOUNT_DELTA modifier targeting the seeded Liquide middelen ledger group
budget-scenarios.spec.ts:516     promoting scenario B to default demotes scenario A in the same action
```

**#943 fails the identical three** — 3 failed / 307 passed on both — and #943 contains *no PHP changes at all*, only the stylelint family bump. Two branches with disjoint diffs failing the same set puts the cause on `development`.

I checked this properly rather than assuming, because this PR *does* touch `BudgetScenarioDefaultPromoter`, `BudgetScenarioModifierGuard` and `BudgetScenarioRegistration` — so "budget specs fail" was a real suspicion, not a formality:

- The earlier run on this branch (**1505**, before those edits) failed only `cbs-submissions`; these three passed. The window that introduced them also pulled in development's #1011 and #998.
- The dead-guard removals cannot change behaviour: `findAll()` is declared `: array` on the canonical contract on **both** `main` and `development`, so `is_array()` on its result was unreachable-as-false.
- The one edit that *does* change behaviour — the DI factory probe — is exonerated by #943, which has none of it.

Those specs are under active repair on `development` right now (fix commits at 00:24 and 02:06 touching exactly these files), so they are someone's work in flight, not mine to take over.

### This PR's own gate is green

```
42 jobs · only E2E + Quality Report red
quality / PHP Quality (phpstan): success      <- the point of the PR
phpstan 0 errors · phpcs 0 errors · 4879 PHPUnit tests green (locally)
```

Worth noting why this is invisible on `development` itself: every recent `development` run is **concurrency-cancelled** by the merge cadence, so the breakage only surfaces on PRs.

This completes the PHPStan 2 migration across all 9 PHP apps.

PR #1069, opened 2026-08-21
test(e2e): scope the modifier row to this test's own scenario

rubenvdlinde:
The last intermittent failure on shillinq `development`. No production code changes — this is the read-back being aimed correctly.

## The defect

The modifier CRUD test located its row by **type alone**:

```js
.filter({ hasText: 'LEDGER_AMOUNT_DELTA' })   ->   .first()
```

Every previous run leaves a `LEDGER_AMOUNT_DELTA` modifier on that index carrying the **same hardcoded** effective date (`2027-09-01`) and amount (`-500000`), so nothing distinguishes this run's row from any other. `.first()` picked an arbitrary one.

When that row's scenario had since been cleaned up, its detail page rendered no data and the final assertion failed with `element(s) not found`.

## The evidence it is a read-back problem, not a write problem

Observed across the six runs available today: **2 failures, both on that same assertion**, never on the create flow above it. The dialog filled, Create enabled, the dialog closed, and the row appeared — every write succeeded. Only the read-back was aimed at the wrong object.

## The fix

The test already creates a uniquely-named scenario via `uniqueSuffix()`, and the index carries a `Scenario` column. So its own row **is** identifiable — the test simply wasn't using it:

```js
.filter({ hasText: 'LEDGER_AMOUNT_DELTA' })
.filter({ hasText: scenarioName })
```

A `toHaveCount(1)` goes in front, so a mis-scoped lookup now fails saying *which way* it was wrong:

- **0 rows** — the `Scenario` column did not render the scenario's name, so the row cannot be identified this way
- **>1 rows** — the fixture leaked

That matters more than the count itself: the previous form could not distinguish "my row is missing" from "I am looking at somebody else's row", which is why it presented as a flake rather than as a defect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/e2e/budget-scenarios.spec.ts
+++ b/tests/e2e/budget-scenarios.spec.ts
@@ -494,6 +494,24 @@ test.describe('budget-scenarios — modifier CRUD reachable (REQ-BSC-008, REQ-BS
 			timeout: 15_000,
 		})
 
+		// Snapshot the index BEFORE creating, so the new row can be identified
+		// by difference rather than by any single column's rendering.
+		//
+		// Scoping by the scenario NAME does not work: the `Scenario` column
+		// renders the `$ref`'s raw id, not the referenced object's name — a
+		// name filter matched 0 rows when this was tried. Scoping by the
+		// modifier's own fields does not work either: the effective date and
+		// amount are hardcoded, so every run's modifier looks identical.
+		//
+		// The row's text DOES differ, because each run's scenario id differs.
+		// Diffing before/after therefore identifies this run's row without
+		// depending on column order, on locale-formatted dates, or on how a
+		// `$ref` happens to be rendered today.
+		const modifierRows = page
+			.locator('table tbody tr')
+			.filter({ hasText: 'LEDGER_AMOUNT_DELTA' })
+		const rowsBefore = new Set(await modifierRows.allTextContents())
+
 		const dialog = await openCreateDialog(page)
 
 		// Labels come from the schema `title` plus the required marker, e.g.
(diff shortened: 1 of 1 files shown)


rubenvdlinde:
Marked draft — this is correct but must not land alone.

The scoping fix works: the count assertion and the diff-based row lookup both passed in CI, so the test now clicks **its own** modifier. What still fails is the final assertion, and that is a genuine product defect, now filed as #1073:

- the detail page fetches its object — `GET …/BudgetScenarioModifier/34d39317-… -> 200`, body contains `"modifierType": "LEDGER_AMOUNT_DELTA"`
- the page renders its title and a **Related (2)** panel with the two `$ref` targets
- **none of the eight declared scalar fields render**

So the old `.first()` behaviour was masking it: the test sometimes clicked *another run's* modifier and passed for reasons unrelated to the object under test. That is why this read as a flake for six runs.

Merging this PR alone would convert an intermittent red into a permanent one on `development` — honest, but it would sit red until #1073 is fixed. Holding it as draft so the two can land together.

Happy to be overruled if you would rather have the deterministic red now: an intermittent failure that misattributes its cause is arguably worse than a red that names it.

rubenvdlinde:
Merging this rather than holding it, and stating the trade-off plainly.

**Why merge now.** The old `.first()` on the modifier type could click *another run's* object, and did — which is how a broken detail page went unnoticed for six runs. A test that passes by reading the wrong object is not testing anything, so leaving it in place keeps a green light that means nothing.

**What it costs.** `development` will show a deterministic failure on this test until #1073 is fixed, instead of an intermittent one. That is the honest state: the page really does not render its scalar fields.

**Why that is the better trade.** An intermittent failure that misattributes its cause is worse than a red that names it. This one now points at #1073, where the mechanism is localised to a single guard (`CnObjectDataWidget` is gated on `currentSchema`; `CnRelatedObjectsWidget` is not) with seven competing hypotheses eliminated.

Revert if you would rather have the quiet amber — but then #1073 loses the only test that demonstrates it.

rubenvdlinde:
Superseded — `development` already carries a better fix.

Commit `183e98ca` (13:20) replaces the row click entirely:

```js
// Its own detail page, addressed by id, must render the type it was
// saved with. Navigating directly also removes the row-click step,
// which was where the wrong-object confusion entered.
await gotoRoute(page, `/begroting/scenario-modifiers/${modifierId}`)
```

That is cleaner than this PR's approach. Mine identified the row by diffing the index before and after the create; addressing the object by id removes the ambiguity at the source rather than working around it. Closing in favour of it.

One thing worth carrying over: the diagnosis in #1073 was gathered from runs **before** `183e98ca` landed, and that commit's own run was cancelled, so it has not yet produced a verdict. If the direct-navigation version passes, #1073 is wrong and should be closed — the failure would have been the wrong object all along. If it fails, #1073 stands and the detail page genuinely does not render its scalar fields. I have said as much on the issue rather than leave a confident claim that the next run might contradict.

Issue #1209, opened 2026-08-24
Eleven components fetch /apps/shillinq/api/openregister/objects/... — a route that does not exist; the SPA catch-all returns 200 + HTML so every read silently renders empty


rubenvdlinde:
Eleven Vue components fetch `/apps/shillinq/api/openregister/objects/...`. **That route does not exist.** `appinfo/routes.php` declares 227 entries and not one matches it, and `\OCA\OpenRegister\AppHost\Routes::standard()` appends a `/{path}` catch-all (`dashboard#catchAll`) as its last route.

So the request is answered by the SPA catch-all: **HTTP 200, `text/html`, ~39 KB of Nextcloud shell.** Not a 404. The caller's `axios.get(...)` resolves successfully, `data.results` is `undefined`, the list renders zero rows, and the component shows its empty state. No error, no console warning, no log line.

## Measured on a live rig

Rig: NC 34 + OpenRegister + shillinq @ `5f382e97`, seeded.

| URL | Result |
|---|---|
| `/apps/shillinq/api/openregister/objects/CommitmentLine?_limit=1` | 200 `text/html`, 39645 B |
| `/apps/shillinq/api/openregister/objects/CommitmentLine/aggregations/committedVsRealisedPerBudgetLine` | 200 `text/html`, 39653 B |
| `/apps/shillinq/api/openregister/objects/Verplichtingsregel?_limit=1` *(a slug that no longer exists)* | 200 `text/html`, 39645 B |
| `/apps/shillinq/api/openregister/objects/ThisSchemaDoesNotExistXyz` *(control)* | 200 `text/html`, 39649 B |
| `/apps/shillinq/api/zzz/nonsense/path` *(control)* | 200 `text/html`, 39653 B |

A real schema, a retired schema, a nonsense schema and a nonsense path are **indistinguishable**.

Against OpenRegister's actual endpoint the same data is there:

- `GET /apps/openregister/api/objects/shillinq/CommitmentLine?_limit=1` → **200, real data** (`VPL-2026-0001`, amounts populated)
- `GET /apps/openregister/api/objects/shillinq/Verplichtingsregel?_limit=1` → **404 `{"message":"Schema not found"}`** — a real answer

## Why the test suite cannot catch it

`tests/e2e/budget-line-commitments.spec.ts` has two tests:

[...]

PR #1211, opened 2026-08-24
refactor(commitment)!: anglicise the commitment enum values, with two migrations

rubenvdlinde:
## What

Tranche 2 of the shillinq commitment anglicisation. Tranche 1 (#1196) moved the schema **names**; this moves the **values** stored inside them.

| Property | Dutch → English |
|---|---|
| `Commitment.status` | `in_goedkeuring`→`in_approval`, `aangegaan`→`committed`, `deels_geleverd`→`partially_delivered`, `deels_gefactureerd`→`partially_invoiced`, `deels_betaald`→`partially_paid`, `afgesloten`→`closed`, `geannuleerd`→`cancelled` |
| `Commitment.kind` | `inkooporder`→`purchase_order`, `arbeidscontract`→`employment_contract`, `subsidiebeschikking`→`grant_decision`, `huurovereenkomst`→`lease_agreement` |
| `Commitment.vat_regime` | `verlegd`→`reverse_charged` |
| `ApprovalStep.role_required` | `budgethouder`→`budget_holder`, `teamleider`→`team_lead`, `directeur`→`director`, `college`→`municipal_executive` |

Every value was verified unique to `bookkeeping-verplichtingenadministratie.json` across the whole register first. That matters because `RenameDutchValueDecisions` is keyed by **property name**, not by schema — a bucket under `status` applies register-wide, so a value shared with another fragment would migrate that fragment's stored data while its schema still declared the Dutch enum.

## `ApprovalStep.status` is deliberately left wholly Dutch

Two of its values — `in_behandeling` and `afgewezen` — are also live in `bookkeeping-rechtmatigheidsverantwoording.json` (`Rechtmatigheidstoets.status`, `Rechtmatigheidsbevinding.status`). Migrating them would desync that cluster's data from its schema.

Renaming only the *other* three would leave the enum half-Dutch — the exact "two vocabularies at once" defect this programme exists to remove. So the whole enum waits for the Rechtmatigheid cluster. A `_note` in the fragment records why.

## The second migration, and why the first could not cover it

`Mandate.kind_commitment` holds the **same four** `Commitment.kind` values, but as a **JSON array** — "this mandate covers these kinds of commitment".

`ValueMigrationPort::rewrite()` is `UPDATE t SET c = ? WHERE c = ?` — an equality match on the whole cell. Against a stored `["inkooporder","leasing"]` it matches nothing and returns `0`.

**Adding a `kind_commitment` bucket would have been worse than omitting one**: the step would report success having migrated no mandate at all, and the entry would make it look covered.

Left unmigrated, `MandateEnforcer::mandateApplies()` does

```php
in_array($kind, $mandate['kind_commitment'], true)
```

with an English `$kind` against stored Dutch members — so the mandate stops matching. **Nothing errors**: an empty list already means "any kind", so a non-empty Dutch list simply fails every comparison. A commitment would be routed to approval, or refused, while the mandate that should have authorised it sits right there in the register.

New repair step **`RenameMandateCommitmentKinds`** does the read-modify-write the port cannot: decode, map the four values, re-encode.

- **Idempotent** — an already-English list returns "no change", so a re-run writes nothing
- **List-only** — a null, scalar or object is skipped, never coerced
- **Preserves unmapped members verbatim** (`leasing`, `other` survive beside a translated value)
- **Register-scoped** and **fail-soft** (a throwing repair step aborts `occ upgrade`)
- Registered **after** `RenameDutchValues`, which moves the scalar column this list mirrors

## Tests

Nine unit tests, including:

- the **ordering** against `info.xml` — a later reshuffle would otherwise reintroduce the bug with no code change
- a test pinning **every migration target against the enum the fragment declares** — if the two halves drift, stored data migrates to a value the schema rejects

The map test was confirmed to **fail (4 failures)** with one entry removed, so it is not vacuous.

One behaviour worth naming: `{"0":"inkooporder"}` and `["inkooporder"]` decode identically under `json_decode($raw, true)`, so no guard can separate them. Such a cell is normalised to a real JSON list — the shape OpenRegister writes anyway — and a test records that as a decision rather than an accident.

## Verified

- `phpunit` — **4950 tests, 46681 assertions, 0 failures** (1 skip, pre-existing and present on `development`)
- `phpcs`, `phpmd`, `psalm`, `phpstan` — all pass
- **All 15 JS validators** pass, including `check:l10n-js` — the check that caught the stale browser catalogues after tranche 1 (#1208)


rubenvdlinde:
## Verified against a real database — and it caught a bug the unit tests could not

Ran this branch on a live rig (NC 34 + OpenRegister + shillinq, Postgres) that happened to hold **two real mandates with Dutch values already stored**:

```
 _id | kind_commitment
-----+------------------------------------------------------------------
   1 | ["inkooporder", "frameworkAgreement"]
   2 | ["inkooporder", "frameworkAgreement", "arbeidscontract",
      |  "huurovereenkomst", "leasing"]
```

`occ maintenance:repair --include-expensive` printed:

```
- RenameMandateCommitmentKinds: rewrote 0 mandate row(s).
```

**Zero, over two rows that plainly needed migrating.**

### Why

The step addressed the shard table as `SELECT id, …` / `WHERE id = ?`. **There is no `id` column** — OpenRegister prefixes every control column of a per-schema shard table with an underscore (`_id`, `_uuid`, `_schema`, `_register`, …) and leaves only the schema's own properties bare.

The SELECT threw, my own fail-soft catch swallowed it, and the step reported `rewrote 0` — which is *also* exactly what a correctly working step prints on an instance with nothing left to migrate. The two outcomes are indistinguishable in the output. Same silent-success shape as #1209 and #1216, this time in my own migration.

**The unit tests could not see it.** A mocked `IResult` returns whatever key the test wrote, so `['id' => 7]` passed happily against SQL Postgres would have rejected. A fixture agreeing with the code it tests is not evidence.

### After the fix

```
- RenameMandateCommitmentKinds: rewrote 2 mandate row(s).

 _id | kind_commitment
-----+------------------------------------------------------------------
   1 | ["purchase_order", "frameworkAgreement"]
   2 | ["purchase_order", "frameworkAgreement", "employment_contract",
      |  "lease_agreement", "leasing"]
```

- Dutch values translated ✅
- **Unmapped members survived verbatim** — `frameworkAgreement`, `leasing`, `other` ✅
- An immediate re-run printed `rewrote 0` — **idempotent against real data**, not just against a mock ✅
- `RenameCommitmentSchemas` (tranche 1) printed `renamed 0 of 7` on this rig, correct: it was installed post-rename, so there were no Dutch slugs left

A new test asserts the SQL **text** contains `SELECT _id, ` and `WHERE _id = ?` and does *not* contain the bare forms. Asserting the string is the only way a unit test can hold this — the column name is a fact about the database, and every mock in the suite is free to disagree with it.

⚠️ Note for anyone testing repair steps by hand: **restart the container after editing** — opcache serves the stale class and the first run will show the old behaviour.

